PR #16, opened 2019-10-16
Get MPI_INFO_NULL from eckit

benjaminmenetrier:
In this branch, I have added a method for the eckit communicator that provides the `MPI_INFO_NULL` value for Fortran applications. This method is then called in our fckit fork.

`MPI_INFO_NULL` value is required for parallel NetCDF calls (argument `info` in [nf90_create](https://www.unidata.ucar.edu/software/netcdf/fortran/docs/f90_datasets.html#f90-nf90_create)), and it is compiler-dependent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,7 +200,7 @@ if( HAVE_ECKIT_CMD)
   list( APPEND ECKIT_LIBRARIES eckit_cmd )
 endif()
 
-if( HAVE_ECKIT_MPI )
+if( HAVE_MPI )
   list( APPEND ECKIT_LIBRARIES eckit_mpi )
 endif()
 
--- a/src/eckit/container/kdtree/KDNode.h
+++ b/src/eckit/container/kdtree/KDNode.h
@@ -45,6 +45,8 @@ public:
 
     static KDNode<Traits>* insert(Alloc& a, const Value& value, KDNode<Traits>* node, int depth = 0);
 
+    /// Return the axis along which this node is split.
+    size_t axis() const { return axis_; }
 
 public:
     void nearestNeighbourX(Alloc& a,const Point& p, Node*& best, double& max, int depth);
--- a/src/eckit/mpi/Comm.h
+++ b/src/eckit/mpi/Comm.h
@@ -110,6 +110,8 @@ public:  // methods
 
     virtual int anyTag() const = 0;
 
+    virtual int infoNull() const = 0;
+
     virtual Status status() const = 0;
 
     virtual Request request(int) const = 0;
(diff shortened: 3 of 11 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=16) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=16) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/eckit?pullRequest=16) it.</sub>

wdeconinck:
Thank you @benjaminmenetrier for this contribution.
CC. @tlmquintino 
It seems we need to review deeper the ```eckit::mpi``` approach. Within Fortran the MPI_Comm and MPI_Info types are integer values. But say you require the C-API of NetCDF, then you would need access to the actual private MPI_Comm object and the actual MPI_Info object.
We need a way to access this. It likely means we will have to do something like
```c++
auto info_null = dynamic_cast< eckit::mpi::Parallel::Info >( eckit::mpi::comm().infoNull() );
auto mpi_info = info_null.MPI_Info();
```
And perhaps something similar for the MPI_Comm:
```c++
auto comm = dynamic_cast< eckit::mpi::Parallel >( eckit::mpi::comm() );
auto mpi_comm = comm.MPI_Comm();
```
So that then these can be used for the NetCDF call [nc_create_par](https://www.unidata.ucar.edu/software/netcdf/docs/group__datasets.html#ga5e03371d14dfac165734b3d0d2c3adc6):
```
nc_create_par( filepath, cmode, mpi_comm, mpi_info, ncdip );
```
All the above is currently not implemented. We will internally look into this and agree on an API. Once that is done, the fckit_mpi interface which you are undoubtedly interested in can be created.

You would also need check if you are actually using the Parallel backend or the Serial one, which would lead to the full proof of concept example
```c++
if( dynamic_cast<eckit::mpi::Parallel*>( &eckit::mpi::comm() ) ) {
  nc_create_par( 
    filepath,
    cmode, 
    dynamic_cast< eckit::mpi::Parallel& >( eckit::mpi::comm() ).MPI_Comm(),
    dynamic_cast< eckit::mpi::Parallel::Info& >( eckit::mpi::comm().infoNull() ).MPI_Info(),
    ncdip );
}
else {
  nc_create( filepath, cmode, ncdip );
}

benjaminmenetrier:
Hi @wdeconinck, you are right, my approach is not completely generic and should be improved. Let me know if I can help.

wdeconinck:
@benjaminmenetrier We will have some time next week to discuss this PR. Sorry for the delay.
But FYI, please do not make a PR from the develop branch in the fork as this is then no longer a fixed PR. As you see more commits have been added unintentionally

tlmquintino:
This is solved in a different way in the upcoming release 1.9.0

PR #17, opened 2019-11-25
Fix missing eckit_mpi.so in ECKIT_LIBRARIES

markjolah:
`ecbuild_add_option(FEATURE MPI ...)` creates variable `HAVE_MPI` but this conditional checks `HAVE_ECKIT_MPI` which is never set.  With this patch, `${CMAKE_INSTALL_PREFIX}/lib/cmake/eckit/eckit-config.cmake` now correctly includes `eckit_mpi` in `ECKIT_LIBRARIES` when MPI is found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,7 +205,7 @@ endif()
 # export package info
 
 set( ECKIT_INCLUDE_DIRS   ${CMAKE_CURRENT_SOURCE_DIR}/src ${CMAKE_CURRENT_BINARY_DIR}/src )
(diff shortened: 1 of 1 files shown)


wdeconinck:
Thank you for your contribution. In fact it seems that the library eckit_mpi is *always* available. The "HAVE_MPI" variable controls whether the "Parallel" (mpi) implementation is compiled in or not.

I would rather remove the if-statement altogether and have "eckit_mpi" be part of ECKIT_LIBRARIES always.

markjolah:
OK, I see.  How is this?

wdeconinck:
Thanks for fixing! I am happy with this. @tlmquintino what about you?

tlmquintino:
I do agree with the proposed changes.

PR #19, opened 2020-01-10
Extend KDNode with a public method returning the node's splitting axis

wsmigaj:
Addition of this method would make it possible to write custom visitors traversing a kd-tree and performing spatial queries different from the ones supported natively by the ``KDNode`` class (i.e. nearest neighbour search and retrieval of points lying in a sphere). For example, queries checking whether any point in a kd-tree lies within a volume such as a cuboid, ellipsoid or cylinder.

The parent ``SPNode`` class already gives clients read-only access to its children nodes, point and payload, so I think making the splitting axis publicly accessible too is consistent with the rest of the interface.

This change has been accepted into the JCSDA fork of ``eckit`` (https://github.com/JCSDA/eckit/pull/8).


tlmquintino:
Thanks for the pull request.
I think the change is ok, but I would like to see that unit test expanded to cover this usage.
Feel free to add a helper class there that exercises the need for this method to be public.
This will stop a future refactoring from removing the method or renaming it.

wsmigaj:
That's a good point. I'll update the test suite so that this method is tested.

tlmquintino:
@wdeconinck I think you can go ahed with the merge

wdeconinck:
Merged

PR #29, opened 2020-05-21
Add waitall & waitany MPI bindings

pmarguinaud:


src/eckit/mpi/Serial.cc
```diff
@@ -192,6 +192,15 @@ Status Serial::wait(Request& req) const {
...
 
+std::vector<Status> Serial::waitall (std::vector<Request> & rq) const {
+    throw NotImplemented ("Serial::waitall", Here ());
+}
+
+Status Serial::waitany(std::vector<Request> &, int &) const
+{
+    throw NotImplemented ("Serial::waitany", Here ());
+}
+
```
wdeconinck, line 203: Both `Serial::waitall()` and `Serial::waitany()` could be implemented simply by making using of the existing `wait()`.
In this serial implementation of isend/ireceive, internal buffers are being kept alive for each request, and the copy will only happen upon the wait.

pmarguinaud: Hello Willem,

I do not know the Serial class; are you sure that depending on the order we process the request list, this could not lead to a deadlock ?

Philippe

pmarguinaud: Hello Willem,

I have looked at the `SerialRequest` class. Requests of this kind have a `test` method which always returns `true`; this implies that no deadlock can ever happen, but that it also hard to implement decently the `waitany` method : if `waitany` is called multiple times on a vector of requests, then it is not possible to pick a request and decide that is has completed (because all of them have already completed).

Another possibility would be to call wait for all requests and stick to the standard which says that : 

> If all of the requests are MPI_REQUEST_NULL, then indx is returned as MPI_UNDEFINED, and status is returned as an empty status. 

But if we do that, then code such as : 

```
    int count = rqr.size (); 
    while (count > 0)
     {   
       int ireq = -1; 
       mpi::Status st = comm.waitany (rqr, ireq); 
       EXPECT (st.error () == 0); 
       count--;
     }   
```
is guaranteed to deadlock, unless the user tests `ireq == MPI_UNDEFINED`. 

The other issue is that MPI_UNDEFINED is not available in Serial.cc (compiled without MPI).

What should I do here ?

wdeconinck: I see the problem now, but only for waitany, right?

Would it be sufficient if there was a mechanism to mark requests as completed?

pmarguinaud: Yes, you are right, for waitall, it is sufficient to loop on the list of requests, I did it and it works.

But for waitany, we need to to return a different ireq each time it is called, which means some extra information is required; maybe the solution you propose will work, I do not know the Serial class well enough.

Philippe

wdeconinck: I can take your branch and figure out a solution to make waitany work in Serial.

FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=29) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=29) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: pmarguinaud<br/>:x: wdeconinck<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/eckit?pullRequest=29) it.</sub>

wdeconinck:
I noticed a test was missing for the Serial backend.
I added that, which is basically running the same test twice, once with MPI (comm.size()==4), and once without MPI.
I of course then needed to then implement`Serial::waitany` and `Serial::waitall`, which seems to pass now.

pmarguinaud:
Thank you Willem.

wdeconinck:
This capability has been merged in develop and master and available in tagged version 1.12.0 see 548cc0a17dadc4f00df694e7f348ece9c8294709

PR #42, opened 2022-02-11
YAML parser: handle block style quoted associative array key

matthewrmshin:
This change ensures that quoted associative array entry keys in _block style_ are parsed correctly by the YAML parser. For example, consider the data where we have to use number-like strings as associative array keys. We have to quote the numbers like this:

```yaml
# block style
"10": ten
"20": twenty
```

which should be equivalent to:

```yaml
# flow style
{"10": "ten", "20": "twenty"}
```

The parser did not parse the _block style_ version correctly before this change.

### DIFF
--- a/src/eckit/parser/YAMLParser.cc
+++ b/src/eckit/parser/YAMLParser.cc
@@ -193,10 +193,6 @@ struct YAMLItemKey : public YAMLItem {
         YAMLItem(item->indent_, item->value_) {
 
         YAMLItemLock lock(item);  // Trigger deletion
-
-        std::string v(value_);
-        ASSERT(v.size());
-        value_ = toValue(v.substr(0, v.size() - 1));
     }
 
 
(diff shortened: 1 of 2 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=42) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=42) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/eckit?pullRequest=42) it.</sub>

tlmquintino:
Thanks for the contribution and for including a unit test.

PR #43, opened 2022-03-09
Add verbose output stream

neill-b:
### Description

The debug output stream is useful to track the progress of an application and check the values of quantities used in calculations. However, we have found several cases in which the debug stream can produce an excessive amount of output. This can happen when, for example, it is placed inside a for loop. In a simple unit test there may only be a few lines of output, but in a full operational test that could suddenly turn into thousands of essentially redundant lines.

### Requirements

Produce a new verbose output stream.  This will be used for normal diagnostic messages, but will be slightly more verbose than the default channel.  It can be switched on by setting the environment variable MAIN_VERBOSE=1.  The verbose channel will also be produced if MAIN_DEBUG=1, as we are thinking of a hierarchy of output levels.  Thus, the debug output stream can be kept for full debugging output.

### Dependencies

There is a corresponding fckit change: https://github.com/ecmwf/fckit/pull/19.


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=43) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/eckit?pullRequest=43) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/eckit?pullRequest=43) it.</sub>

tlmquintino:
Hi @neill-b 

Thanks for your interest in eckit. We've been thinking and discussing this internally. 

Addressing verbosity is not the same as addressing the intent of the channel. 
Verbosity controls the content (information) sent through the channel, whereas the channel controls where and how the streams (carrying the content) are sent to (eg. file, stdout, etc)

So I think this is mixing the concerns/responsibility of the channel with the stream content management. Therefore I don't think this is the way to attack your problem.

May I propose a solution that you can implement on the application side?
You can wrap the INFO or DEBUG channel with an object of your own and only truly pass the stream onwards when the verbosity flag is set. You could even go further and implement multiple verbosity levels, from quiet to full verbosity, but this may require that your application then uses the level in the output lines.

neill-b:
Thanks @tlmquintino for the feedback.  We accept that streams and verbosity are two separate issues.  We are attempting to come up with a strategy about how to address this within JEDI.

PR #46, opened 2022-07-07
Improve YAML integer/float parse logic

matthewrmshin:
Improve compatibility with YAML ~1.1~ 1.2 for number parsing logic:
* ~Handle base 2 integer.~
* ~Handle base 60 integer/float.~
* Handle float `Inf` and `NaN`.
* Recognise `~` as a `NULL` value.
* Improve base 16 integer matching logic. (`[a-f]` not `[a-z]`!)
* Base 8 integer expects `0o` (Zero Oh) prefix.
  * `0o123` is a base 8 integer. (`83` in base 10.)
  * `0123` is `123` in base 10.
* ~Unquoted digits-only value with a leading `0` (zero), but with the characters `8` and/or `9` will now be considered a string instead of a number.~
* Expect quotes around digit-only string values to avoid misinterpretation as integers.
* Handle and ignore `_` (underscore) characters in numbers.

See:
* https://yaml.org/spec/1.2.2/#103-core-schema
* ~https://yaml.org/type/int.html~
* ~https://yaml.org/type/float.html~


matthewrmshin:
@simondsmart @danovaro Any chance that this PR can be assigned for review? Cheers. 👍🏼

tlmquintino:
> @simondsmart @danovaro Any chance that this PR can be assigned for review? Cheers. 👍🏼

Thanks for providing this PR. Looks good and we are starting to review it.

matthewrmshin:
@tlmquintino Thanks!

src/eckit/parser/YAMLParser.cc
```diff
@@ -443,27 +479,30 @@ static Value toValue(const std::string& s) {
...
+                        str = str.substr(0, 1) + str.substr(3);
+                    }
+                    return Value(strtol(str.c_str(), 0, 2));
                 }
 
-                if (s.length() > 2 && s[1] == 'x' && hex.match(s)) {
-                    return Value(strtol(s.c_str(), 0, 0));
+                if (integer8.match(s)) {
+                    std::string str = ignoreChar(s, '_');
+                    return Value(strtol(str.c_str(), 0, 0));
```
simondsmart, line 498: Should we not be being explicit about base 8 here, given we have already tested for it?

src/eckit/parser/YAMLParser.cc
```diff
@@ -443,27 +479,30 @@ static Value toValue(const std::string& s) {
...
                 }
 
-                if (integer.match(s)) {
-                    long long d = Translator<std::string, long long>()(s);
-                    return Value(d);
+                if (integer16.match(s)) {
+                    std::string str = ignoreChar(s, '_');
+                    return Value(strtol(str.c_str(), 0, 0));
```
simondsmart, line 503: Similarly, should we not specify base 16 in strtol

src/eckit/parser/YAMLParser.cc
```diff
@@ -443,27 +479,30 @@ static Value toValue(const std::string& s) {
...
         // This is because checking regex is very slow
         switch (s[0]) {
+            case '-':
+            case '+':
             case '0':
 
-                if (octal.match(s)) {
-                    return Value(strtol(s.c_str(), 0, 0));
+                if (integer2.match(s)) {
+                    std::string str = ignoreChar(s, '_');
```
simondsmart, line 487: Typically ignoreChars will be a nop (most of the numbers parsed will not be containing '_'). This introduces an annoying memory copy here. Is it possible to avoid creating a new string except in cases where it is needed?

src/eckit/parser/YAMLParser.cc
```diff
@@ -425,12 +426,47 @@ Value YAMLParser::parseNumber() {
...
 
 
+std::string ignoreChar(const std::string& s, const char ignore) {
+    std::string str = s;
+    str.erase(remove(str.begin(), str.end(), ignore), str.end());
+    return str;
+}
+
+
+long long parseSexagesimal(const std::string& s) {
```
simondsmart, line 436: Do we want to support Sexagesimal literals? They have been removed in YAML 1.2 IIRC, and cause really awkward bugs if you ever want to have (say) MAC addresses in a YAML... (i.e. they can misparse ONLY if they contain no hex digits A-F)

src/eckit/parser/YAMLParser.cc
```diff
@@ -425,12 +426,47 @@ Value YAMLParser::parseNumber() {
 }
```
simondsmart, line 428: By preference would be to put helper functions into an anonymous namespace. And at the very least they should be declared static so they don't leak into wider linkage.

tests/parser/2.19.yaml
```diff
@@ -1,5 +1,19 @@
-canonical: 12345
-decimal: +12,345
-sexagesimal: 3:25:45
-octal: 014
-hexadecimal: 0xC
+base2: 0b11010100001011
+base8: 032413
```
simondsmart, line 2: Can we test some of the awkward cases?

e.g. 038123 (i.e. it should be a string(?))

i.e. what are the edge cases that these rules might fail in. The rules have become much more complex, with significantly more complex regexes, so there are quite a lot more edge cases to check.

simondsmart (changes requested):
Overall, very happy with this. It looks like a good improvement!

Just a couple of questions in there that I'd like you to look at. Also given the number of possible edge cases in the regexes involved, I'd like to have some more test cases probing the corners of the pattern matching.

Otherwise, we should be able to merge it shortly.

matthewrmshin:
@simondsmart Thanks for the detailed review. 👍🏼  I'll make the recommended adjustments later today.

On YAML 1.1 to YAML 1.2:
* You are right about base 60 integer/float (and base 2 integer!) being withdrawn.
* However, base 8 integer representation has also changed. A YAML 1.2 parser will only recognise an unquoted value as a base 8 integer if it starts with `0o` (zero oh). An unquoted digit-only value is now a base 10 integer.
  * So `0o1234` is an integer (`668` in base 10).
  * So is `01234` which is now just `1234`.
  * And digit-only strings should all be quoted.

Happy to adjust the code to be YAML 1.2 compliant (instead of YAML 1.1 compliant). What do you think? 💡

src/eckit/parser/YAMLParser.cc
matthewrmshin: Done.

src/eckit/parser/YAMLParser.cc
matthewrmshin: Done.

src/eckit/parser/YAMLParser.cc
matthewrmshin: I have made a change to `ignoreChar`. Let me know if it does what you want here or not.

src/eckit/parser/YAMLParser.cc
matthewrmshin: I have removed this as commented below.

tests/parser/2.19.yaml
matthewrmshin: There should now be some string cases later in the file.

src/eckit/parser/YAMLParser.cc
```diff
@@ -24,6 +25,7 @@
 
 namespace eckit {
 
+static std::string ignoreChar(const std::string& s, const char ignore);
```
simondsmart, line 28: I'm not sure that we need a forward declaration of this?

src/eckit/parser/YAMLParser.cc
```diff
@@ -425,12 +427,26 @@ Value YAMLParser::parseNumber() {
...
 
+static std::string ignoreChar(const std::string& s, const char ignore) {
+    if (s.find(ignore) == std::string::npos) {
+        return s;
+    }
+    std::string str = s;
+    str.erase(remove(str.begin(), str.end(), ignore), str.end());
+    return str;
+}
+
```
simondsmart, line 438: This still introduces an unnecessary copy. We can note that at the point that ignoreChar is called, the original string is never required again. So we can take ownership of the string here.

static std::string ignoreChar(std::string&& s, const char ignore) {
    std::string str(std::move(s));
    s.erase(std::remove(s.begin(), s.end(), ignore), s.end());
    return s;
}

or you could make this simpler by just being a wrapper around erase

static void eraseChar(std::string& s, const char ignore) {
    s.erase(std::remove(s.begin(), s.end(), ignore), s.end());
    return s;
}

To be honest, if I were writing it myself, I might just use erase/remove myself directly, in the calling code...

simondsmart:
Thanks @matthewrmshin - this is looking very good. Pretty much happy to merge.

I've made one comment about an unnecessary string copy. I also might consider supporting both the 0XXXXX and 0oXXXXX patterns for octal numbers, just to reduce the possible surprise factor for any users.

src/eckit/parser/YAMLParser.cc
matthewrmshin: @simondsmart Thanks for the detailed feedback again. I'll take the recommendation on ignoreChar and inline it.

matthewrmshin:
The problem with the `0123` syntax is the explicit change between YAML 1.1 and YAML 1.2. In YAML 1.1, it is a base 8 integer. In YAML 1.2, it is a base 10 integer. See for example:
https://yaml.org/spec/1.2.2/ext/changes/#changes-in-version-12-revision-120-2009-07-21

So we can only choose one behaviour. I have currently chosen base 10 for better alignment with YAML 1.2. What do you think?

matthewrmshin:
@simondsmart Many thanks for reviewing and merging my change. 👍🏼

PR #60, opened 2023-02-28
ECKIT-600 Always define preprocessor definitions, with value 0 or 1

wdeconinck:
Copy of [ECKIT-600](https://jira.ecmwf.int/browse/ECKIT-600)

Change preprocessor definitions like `'eckit_HAVE_'` to be always defined with value 0 or 1 is safer and more convenient.
Always being defined means that we use `#if` instead of `#ifdef`. That allows compiler to detect typo in the variable name.
The variable can also be used as boolean.
This is a followup of PR associated with [ECKIT-599](https://jira.ecmwf.int/browse/ECKIT-599)

This had already been discussed in above mentioned PR and agreed by @tlmquintino 
See here https://git.ecmwf.int/projects/ECSDK/repos/eckit/pull-requests/320/diff?commentId=27609#src/eckit/linalg/LinearAlgebraDense.cc?t=33

### DIFF
--- a/src/eckit/CMakeLists.txt
+++ b/src/eckit/CMakeLists.txt
@@ -3,15 +3,15 @@
 test_big_endian( _BIG_ENDIAN )
 
 if( _BIG_ENDIAN )
-        set( ECKIT_BIG_ENDIAN    1 )
-        set( ECKIT_LITTLE_ENDIAN 0 )
+        set( eckit_BIG_ENDIAN    1 )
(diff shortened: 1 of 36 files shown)


wdeconinck:
Rebased to develop again. Current downstream-ci-hpc and private-downstream-ci-hpc seems to work. Does that cover everything, @danovaro ?

danovaro (approved):
tested as part of mars-client, mars-server and pgen bundles.

PR #68, opened 2023-04-26
Enable convertSphericalToCartesian to handle latitudes across poles

fmahebert:
The atlas library generates points with latitudes outside the usual [-90°, 90°] interval, for example halo points on a regular Gaussian grid. Passing these points through eckit's `convertSphericalToCartesian` throws an exception because the code expects latitudes within the usual interval.

This PR preserves the current default behavior, but adds the option to renormalise points with latitudes "across the pole".

(I'd be happy to update the PR to remove the bool and always apply the renormalisation, if that is the preferred behavior. This would enable the change to propagate transparently to atlas without adding the bool, but would affect users who relied on the exception for points outside [-90,90]).

### DIFF
--- a/src/eckit/geometry/Sphere.cc
+++ b/src/eckit/geometry/Sphere.cc
@@ -12,9 +12,7 @@
 
 #include <algorithm>
 #include <cmath>
-#include <ios>
 #include <limits>
-#include <sstream>
 
 #include "eckit/exception/Exceptions.h"
 #include "eckit/geometry/CoordinateHelpers.h"
@@ -33,17 +31,13 @@ static const double radians_to_degrees = 180. * M_1_PI;
 
 static const double degrees_to_radians = M_PI / 180.;
 
-static constexpr std::streamsize max_digits10 = std::numeric_limits<double>::max_digits10;
-
 inline double squared(double x) {
     return x * x;
 }
 
 //----------------------------------------------------------------------------------------------------------------------
 
-double Sphere::centralAngle(const Point2& Alonlat, const Point2& Blonlat) {
-    using namespace std;
-
+double Sphere::centralAngle(const Point2& Alonlat, const Point2& Blonlat, bool normalise_angle) {
     /*
      * Δσ = atan( ((cos(ϕ2) * sin(Δλ))^2 + (cos(ϕ1) * sin(ϕ2) - sin(ϕ1) * cos(ϕ2) * cos(Δλ))^2) /
      *            (sin(ϕ1) * sin(ϕ2) + cos(ϕ1) * cos(ϕ2) * cos(Δλ)) )
@@ -60,6 +54,11 @@ double Sphere::centralAngle(const Point2& Alonlat, const Point2& Blonlat) {
      * }
      */
 
+    if (!normalise_angle) {
+        assert_latitude_range(Alonlat[1]);
+        assert_latitude_range(Blonlat[1]);
+    }
+
     const Point2 alonlat = canonicaliseOnSphere(Alonlat);
     const Point2 blonlat = canonicaliseOnSphere(Blonlat);
 
@@ -67,15 +66,17 @@ double Sphere::centralAngle(const Point2& Alonlat, const Point2& Blonlat) {
     const double phi2   = degrees_to_radians * blonlat[1];
     const double lambda = degrees_to_radians * (blonlat[0] - alonlat[0]);
 
-    const double cos_phi1   = cos(phi1);
-    const double sin_phi1   = sin(phi1);
-    const double cos_phi2   = cos(phi2);
-    const double sin_phi2   = sin(phi2);
-    const double cos_lambda = cos(lambda);
-    const double sin_lambda = sin(lambda);
+    const double cos_phi1   = std::cos(phi1);
+    const double sin_phi1   = std::sin(phi1);
+    const double cos_phi2   = std::cos(phi2);
+    const double sin_phi2   = std::sin(phi2);
+    const double cos_lambda = std::cos(lambda);
+    const double sin_lambda = std::sin(lambda);
 
-    const double angle = atan2(sqrt(squared(cos_phi2 * sin_lambda) + squared(cos_phi1 * sin_phi2 - sin_phi1 * cos_phi2 * cos_lambda)),
-                               sin_phi1 * sin_phi2 + cos_phi1 * cos_phi2 * cos_lambda);
+    const double angle = std::atan2(std::sqrt(squared(cos_phi2 * sin_lambda)
+                                              + squared(cos_phi1 * sin_phi2
+                                                        - sin_phi1 * cos_phi2 * cos_lambda)),
+                                    sin_phi1 * sin_phi2 + cos_phi1 * cos_phi2 * cos_lambda);
 
     if (types::is_approximately_equal(angle, 0.)) {
         return 0.;
@@ -102,7 +103,7 @@ double Sphere::centralAngle(const double& radius, const Point3& A, const Point3&
 }
 
 double Sphere::distance(const double& radius, const Point2& Alonlat, const Point2& Blonlat) {
-    return radius * centralAngle(Alonlat, Blonlat);
+    return radius * centralAngle(Alonlat, Blonlat, true);
 }
 
 double Sphere::distance(const double& radius, const Point3& A, const Point3& B) {
@@ -155,7 +156,8 @@ void Sphere::greatCircleLongitudeGivenLatitude(const Point2& Alonlat, const Poin
     Clon2 = lon.size() > 1 ? lon[1] : std::numeric_limits<double>::signaling_NaN();
 }
 
-void Sphere::convertSphericalToCartesian(const double& radius, const Point2& Alonlat, Point3& B, double height) {
+void Sphere::convertSphericalToCartesian(
+    const double& radius, const Point2& Alonlat, Point3& B, double height, bool normalise_angle) {
     ASSERT(radius > 0.);
 
     /*
@@ -170,6 +172,10 @@ void Sphere::convertSphericalToCartesian(const double& radius, const Point2& Alo
      * These three conditionings combined project very accurately to the sphere
      * poles and quadrants.
      */
+
+    if (!normalise_angle) {
+        assert_latitude_range(Alonlat[1]);
+    }
 
     const Point2 alonlat = canonicaliseOnSphere(Alonlat, -180.);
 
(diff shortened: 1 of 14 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=68) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=68) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/eckit?pullRequest=68) it.</sub>

fmahebert:
@wdeconinck, pinging you as requested.

fmahebert:
> Regarding the boolean flag, I would remove it, as well as the exception throwing.

Done. The second commit halfway undoes the first commit, so this'll be easier to review as a total diff.

fmahebert:
For increased safety, I could handle cases where latitudes are outside [-270,270] ("across two poles") either by throwing an exception or by generalizing the normalization a bit more. Opinions on this?

fmahebert:
I've made the normalisation of the latitudes more robust and added and a new test case. I think I'm satisfied, let me know what you think.

src/eckit/geometry/Sphere.cc
```diff
@@ -197,9 +190,23 @@ void Sphere::convertSphericalToCartesian(const double& radius, const Point2& Alo
...
+     * invariant under this transformation, the normalisation procedure is
+     * simplified to only perform the longitude phase shift.
      */
 
-    const double lambda_deg = normalise_longitude(Alonlat[0], -180.);
+    // We normalise the latitude by calling the logically-identical normalise_longitude.
+    const auto& normalise_latitude = normalise_longitude;
+    const bool lat_across_pole = (normalise_latitude(Alonlat[1], -90.) > 90.);
+
+    const double lambda_deg = normalise_longitude(Alonlat[0] + (lat_across_pole ? 180. : 0.), -180.);
```
MarekWlasak, line 209: Hi - Maybe I am missing something but I can't quite see how you are dealing with the South Pole.

wdeconinck:
```
const bool lat_across_pole = (normalise_latitude(Alonlat[1], -90.) > 90.);
```
does not apply normalisation to the used latitude. It just checks if values are between -90 and +90.
`normalise_latitude(Alonlat[1], -90.)` -> always value between -90. and 270.
So I'm wondering, how is this different from
```
const bool lat_across_pole=(Alonlat[1] > 90. || Alonlat[1] < -90.);
```

tlmquintino:
I would like @pmaciel to review this as this may have impact to https://github.com/ecmwf/mir

src/eckit/geometry/Sphere.cc
fmahebert: If the input latitude has sufficiently large values that it "wraps around the poles" a few times, then the two are different. This is because a point can have very large latitudes still be in the same hemisphere. As an extreme example, consider `(lon,lat) = (x,370)`; this point is also `(x,10)`, in the same hemisphere, and the coordinates can be "fixed" without needing to shift the longitude.

In this example...
- `(Alonlat[1] > 90. || Alonlat[1] < -90.)` would evaluate to true, even though we do not want to shift the longitude value
- `normalise_latitude(Alonlat[1], -90.)` returns 10, which is in [-90,90], consistent with not editing the longitude

Does that help? Or have I missed something?

fmahebert: @MarekWlasak the statement `normalise_latitude(Alonlat[1], -90.)` means that points "below the south pole" are re-labeled with a latitude "above the north pole"... e.g. `lat=-100 --> lat=260`. Using the `normalise_latitude` function avoids writing complicated branching to achieve the desired behavior.

fmahebert: Do I need to write more documentation? :)

pmaciel:
Hi @MarekWlasak, this is a very welcome change, because it addresses a fundamental issue on the assumptions regarding spherical coordinates: that we normalise the longitude, but not the latitude -- asserting that latitudes are within [-90, 90].

Let me add that this is an improvement, definitely. I have recently written very similar code to handle exactly this case (for a different project, this code returns normalised coordinates):
`        lat = normal(lat, -90.);

        if (lat > 90.) {
            lat = 180. - lat;
            lon += 180.;
        }

        return {lat, lat == -90. || lat == 90. ? 0. : normal(lon, lon_minimum)};
`
However, there are two issues with this PR:

1. Is keeping the assumption [-90 <= lat <= 90 ] (the one we are removing) useful? My opinion (like yours, and @wdeconinck) is to remove this assumption -- but the consequence is that maybe asserting on [-90 <= lat <= 90 ] has to be moved downstream the ECMWF software stack (Atlas and MIR.) This ihas been very useful in the past, for robustness on user requests (in the millions per day) and improving the software stack. For this, we need the input from @tlmquintino and @wdeconinck (before discussion, I'm supporting this change)
2. That this PR is not complete: This is only affecting eckit::geometry::Sphere::convertSphericalToCartesian, but not the method (in the same class) centralAngle. And, in addition, EllipsoidOfRevolution::convertSphericalToCartesian and polygon::LonLatPolygon::contains (this specific one being optional).

Whichever direction we do, this can be brought in quite quickly, as this code is really well tested. So discussion should take place first.

fmahebert:
@pmaciel Thanks for your feedback. I'll be happy to address (2), if the conclusion after discussing (1) is to proceed.

src/eckit/geometry/Sphere.cc
fmahebert: @MarekWlasak can you clarify what discontinuity you're referring to? This is just choosing a consistent "branch cut" of the periodic coordinates before converting to cartesian coordinates... I don't expect anything will be discontinuous in the output space...

fmahebert:
@pmaciel @wdeconinck Is there any update on whether item (1) is satisfactory? I don't want to put in the work for (2) if the PR will be rejected, but if doing (2) would help acceptance, I can do it now, just let me know...

pmaciel:
Hi @fmahebert, sorry to not get back sooner. We decided to proceed with (1), ie. remove these assertions and apply similar changes to the remainder of the code. The other classes/method I listed is not comprehensive, but sufficient for this PR, could you please do these too?

I suggest to rename normalise_longitude to normalise_angle (and remove normalise_latitude).

fmahebert:
@pmaciel Thanks! I've added 3 commits that make the same change to methods `Sphere::centralAngle`, `EllipsoidOfRevolution::convertSphericalToCartesian` and `LonLatPolygon::contains`, and also renames to `normalise_angle`.

I've also added a 4th commit that pulls all this common logic into a helper function — this may be a tad less efficient (a few more computations and stack variables), but should be easier to maintain. Up to you if you want this commit, if not I'll revert it.

pmaciel:
Hi @fmahebert , @ytremolet, sorry for the late reply. Yes, for me this is done -- thank you for the contribution. 

I'm currently working on additional functionality that will supercede this (it will include a PointLonLat type, similar to Atlas/MIR) but in the meantime this is a great addition. We will follow this with changes downstream to protect against wrongly calculated latitudes.

pmaciel:
@fmahebert please note, recently a PR was merged into the develop brach moving the code base to C++17 (as a minimum requirement). This is planned for a release soon. Can you please rebase your PR to an updated develop tree? Thanks! (sorry for the extra work!)

fmahebert:
@pmaciel I've rebased; please let me know if there's anything else.

Your described future developments on this front sound good too :)

pmaciel (approved):

pmaciel:
From internal discussions, so that we can smoothly transition the behaviour, I've added an extra argument (bool normalise_angle = false) where appropriate. This is an API change, to make optional the assertion -90<=lat<=90 (which is still the default ebhaviour), so you might need minor adaptations.

I've made the necesary changes, including updating the new unit tests.

fmahebert:
@pmaciel Sorry for the slower response this time; I've merged your commit and added a tiny test fix as well.

tlmquintino:
For what I can see there is a missing test for the assert / throw, when normalise_angle = false, ie. EXPECT_THROWS(...)

fmahebert:
Added some new test statements for the throw

DJDavies2:
I am getting a compile failure with this:

/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc:388:0: error: unterminated argument list invoking macro "EXPECT_THROWS_AS"
 }
 
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc: In function 'void eckit::test::test_88(std::__cxx11::string&, int&, int)':
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc:270:9: error: 'EXPECT_THROWS_AS' was not declared in this scope
         EXPECT_THROWS_AS(poly.contains({lonmid, 180. - latmid}, eckit::BadValue);
         ^~~~~~~~~~~~~~~~
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc:270:9: error: expected '}' at end of input
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc:270:9: error: expected '}' at end of input
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc: At global scope:
/home/d03/frwd/cylc-run/eckit-68/share/mo-bundle/eckit/tests/geometry/test_polygon.cc:270:9: error: expected '}' at end of input
gmake[2]: *** [eckit/tests/geometry/CMakeFiles/eckit_test_geometry_polygon.dir/test_polygon.cc.o] Error 1
gmake[1]: *** [eckit/tests/geometry/CMakeFiles/eckit_test_geometry_polygon.dir/all] Error 2

This is with gnu 7.3.0.

fmahebert:
> there is a parenthesis missing to be closed

... serves me right for pushing an "easy" fix without compiling! 🙃 

Fixed.

tlmquintino (approved):
Looks ok now

fmahebert:
Thanks @pmaciel @tlmquintino and @wdeconinck for iterating on this!

wdeconinck:
@tlmquintino, @simondsmart, please could we create a tagged release with this API change, so I can ensure this API exists in Atlas with a version check?

danovaro:
@wdeconinck just released with tag 1.24.0

PR #70, opened 2023-05-10
C++17 minimum required standard

pmaciel:
Update minimum required C++ standard to C++17

There are some minor automated code updates, which I checked, to significantly lower the number of warnings (IMHO, the remainder warnings should eventually be addressed.)


simondsmart:
This looks great!

PR #73, opened 2023-05-18
Add MPI on/off to build matrix

figi44:
Adds build configurations to build with and without MPI on the HPC.


figi44:
Need to wait untill runners have MPI installed

PR #76, opened 2023-07-03
Add possibility to add plugin search paths

wdeconinck:
The development of plugins is designed with default following structure:
```
<plugin-path>/share/plugins/manifest.yml
<plugin-path>/lib/lib<plugin>.so
```
This PR creates a function `LibraryManager::addPluginSearchPath( path )`, where `path` is a colon-separated list.
The plugin search paths are then also used for the manifest-scan-path and the dynamic-library-path.

### DIFF
--- a/src/eckit/system/LibraryManager.cc
+++ b/src/eckit/system/LibraryManager.cc
@@ -154,16 +154,30 @@ public:  // methods
         return *(j->second);
     }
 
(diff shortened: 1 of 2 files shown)


tlmquintino (approved):
Looks good and adds flexibility for applications to extend.

PR #83, opened 2023-09-20
ECKIT-620 Enable multi-value options in CmdArgs

marcosbento:
This pull request enables the support for multi-value options in CmdArgs.

A multi-value Option allows collecting multiple argv values, as if these values where positional relative to the option e.g. `--option <command> <name> <value> /path/to/one /path/to/two ... --next`.

The goal is to allow collecting argv values until: 1) there are no more values to collect, either because the end of the list of values was reached or because the next option was found; or, 2) a mandatory+optional number of values has been collected.

### DIFF
--- a/src/eckit/option/CmdArgs.cc
+++ b/src/eckit/option/CmdArgs.cc
@@ -46,57 +46,70 @@ CmdArgs::CmdArgs(std::function<void(const std::string&)> usage, std::vector<Opti
 
 void CmdArgs::init(std::function<void(const std::string&)> usage, int args_count, int minimum_args,
                    bool throw_on_error) {
-    Main& ctx  = Main::instance();
-    tool_      = ctx.name();
-    int argc   = ctx.argc();
-    bool error = false;
+    const Main& ctx = Main::instance();
+    tool_           = ctx.name();
+    int argc        = ctx.argc();
+    bool error      = false;
 
-    std::map<std::string, const option::Option*> opts;
+    using options_map_t = std::map<std::string, const option::Option*>;
 
-    for (std::vector<option::Option*>::const_iterator j = options_.begin(); j != options_.end(); ++j) {
-        if ((*j)->active()) {
-            ASSERT(opts.find((*j)->name()) == opts.end());
-            opts[(*j)->name()] = *j;
-            keys_.insert((*j)->name());
-            (*j)->setDefault(*this);
+    options_map_t opts;
+
+    // Fill in 'keys_' and prepare options map
+    for (Option* j : options_) {
+        if (j->active()) {
+            ASSERT(opts.find(j->name()) == opts.end());
+            keys_.insert(j->name());
+            opts[j->name()] = j;
+            j->setDefault(*this);
         }
     }
 
+    const static std::string prefix = "--";
     Tokenizer parse("=");
+    // Process all options/values in argv, letting each Option collect the necessary entries
     for (int i = 1; i < argc; ++i) {
         std::string a = ctx.argv(i);
-        if (a.size() > 2 && a[0] == '-' && a[1] == '-') {
-            std::vector<std::string> v;
-            parse(a.substr(2), v);
+        if (a.substr(0, prefix.size()) == prefix) { // An Option 'a' is found (starts with '--')!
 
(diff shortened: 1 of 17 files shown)


pmaciel:
Only remaining the pgen tests!

marcosbento:
@pmaciel , @b8raoult 

All downstream checks are now passing. Any chance you can have a look at the changes?

src/eckit/option/CmdArgs.cc
```diff
@@ -44,60 +43,85 @@ CmdArgs::CmdArgs(std::function<void(const std::string&)> usage, std::vector<Opti
     init(usage, args_count, minimum_args, throw_on_error);
 }
 
+namespace {
```
pmaciel, line 46: Consider moving to Tokenizer

src/eckit/option/MultiValueOption.cc
```diff
@@ -0,0 +1,93 @@
...
+#include "eckit/config/Configured.h"
+
+#include "eckit/exception/Exceptions.h"
+
+namespace eckit::option {
+
+MultiValueOption::MultiValueOption(const std::string& name, const std::string& description, size_t n_mandatory_values) :
+    MultiValueOption(name, description, n_mandatory_values, 0) {}
+
+MultiValueOption::MultiValueOption(const std::string& name, const std::string& description, size_t n_mandatory_values,
```
pmaciel, line 26: maybe reuse construcur (so there's only one assertion of the same kind)

src/eckit/option/Option.h
```diff
@@ -17,52 +17,96 @@
...
+template <class T>
+class BaseOption : public Option {
+public:
+    BaseOption(const std::string& name, const std::string& description) :
+        Option(name, description), default_value_{std::nullopt} {};
+    BaseOption(const std::string& name, const std::string& description, const T& default_value_) :
+        Option(name, description), default_value_{std::make_optional(default_value_)} {};
+    ~BaseOption() override = default;
+
+    [[deprecated("Specify the default value(s) when calling the ctor")]]
```
pmaciel, line 86: Hurray!

pmaciel (approved):

tests/option/eckit_test_option_cmdargs.cc
pmaciel, line 1: Try adding this

CASE("test_eckit_option_cmdargs_argument_with_dot") {
    // Set up the parser to accept arguments containing equals as values
    // n.b. Option* are deleted inside CmdArgs.
    std::vector<Option*> options;
    options.push_back(new SimpleOption<std::string>("a.a", ""));
    options.push_back(new SimpleOption<std::string>("b.b.b", ""));
    options.push_back(new SimpleOption<std::string>("c.", ""));
    options.push_back(new SimpleOption<std::string>(".d", ""));

    const char* input[] = {"exe", "--a.a=a", "--b.b.b=b", "--c.=c", "--.d=d"};
    Main::initialise(5, const_cast<char**>(input));

    CmdArgs args(&usage, options, 0, 0, true);
    Log::info() << args << std::endl;

    std::string a;
    EXPECT(args.get("a.a", a));
    EXPECT(a == "a");
    args.get("a", a); Log::info() << "a='" << a << "'" << std::endl;

    std::string b;
    EXPECT(args.get("b.b.b", b));
    EXPECT(b == "b");
    args.get("b", b); Log::info() << "b='" << b << "'" << std::endl;

    std::string c;
    EXPECT(args.get("c.", c));
    EXPECT(c == "c");
    args.get("c", c); Log::info() << "c='" << c << "'" << std::endl;

    std::string d;
    EXPECT(args.get(".d", d));
    EXPECT(d == "d");
    args.get("d", d); Log::info() << "d='" << d << "'" << std::endl;
}

It fails, but it is important to keep track of what works and not (so maybe commented, or an expect fails in CMakeLists.txt, so that we can also track when it does work!)

tlmquintino: Thanks @pmaciel for the awesome review!

marcosbento: The test has been added, and extended as discussed/agreed.

pmaciel:
Hi @simondsmart , @b8raoult , @danovaro , shall we merge this? (state what you think)

PR #85, opened 2023-09-27
Feature/sub hourly

danovaro:


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=85) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=85) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/eckit?pullRequest=85) it.</sub>

src/eckit/types/Time.cc
```diff
@@ -27,80 +30,98 @@ inline void printTime(std::ostream& s, long n) {
...
 }
 
-Time::Time(long seconds) :
+Time::Time(long seconds, bool extended) :
     seconds_(seconds) {
-    if (seconds >= 86400 || seconds < 0) {
-        std::string msg = "Time in seconds must be positive and cannot exceed 86400, seconds: ";
+    if ((seconds >= 86400 && !extended) || seconds < 0) {
+        std::string msg = "Time in seconds must be positive and less than 86400 seconds (24h): ";
         Translator<long, std::string> t;
```
pmaciel, line 37: Convert Translator<long, std::string> to std::to_string

src/eckit/types/Time.cc
```diff
@@ -8,6 +8,9 @@
  * does it submit to any jurisdiction.
  */
 
+#include <regex>
+#include <cmath>
+
 #include "eckit/eckit.h"
 
 #include "eckit/persist/DumpLoad.h"
```
pmaciel, line 16: Remove #include "eckit/utils/Tokenizer.h"

src/eckit/types/Time.cc
```diff
@@ -27,80 +30,98 @@ inline void printTime(std::ostream& s, long n) {
...
-                case 6:
-                    hh = t / 10000;
-                    mm = (t % 10000) / 100;
-                    ss = (t % 10000) % 100;
-                    break;
-                default:
-                    err = true;
-                    break;
+Time::Time(const std::string& s, bool extended) {
+    long ss = 0, mm = 0, hh = 0, dd = 0;
```
pmaciel, line 44: Declarations should be one-per-line

src/eckit/types/Time.h
```diff
@@ -32,9 +32,9 @@ typedef double Second;
 class Time {
```
pmaciel, line 32: Modernise
typedef double Second
to
using Second = double;

src/eckit/types/Time.h
```diff
@@ -32,9 +32,9 @@ typedef double Second;
...
 
 public:  // methods
-    Time(long, long, long);
-    Time(long seconds = 0);
-    Time(const std::string&);
+    Time(long, long, long, bool extended = false);
+    Time(long seconds = 0, bool extended = false);
+    Time(const std::string&, bool extended = false);
```
pmaciel, line 37: Missing the arguments names (since they are not deducible by the type).

To be consistent, it's also missing one constructor
    Time(long minutes, long seconds, bool extended = false);

src/eckit/types/Time.cc
```diff
@@ -27,80 +30,98 @@ inline void printTime(std::ostream& s, long n) {
...
+                }
+            }
+        }
     }
 
-    if (hh >= 24 || mm >= 60 || ss >= 60 || hh < 0 || mm < 0 || ss < 0) {
+    if (mm >= 60 || ss >= 60 || (!extended && (hh >= 24 || dd > 0 || hh < 0 || mm < 0 || ss < 0))) {
         std::string msg = "Wrong input for time: ";
         Translator<long, std::string> t;
+        if (dd>0) {
```
pmaciel, line 121: This isn't consistent with the check above, for ss<86400. (eg, line 35 but there are more). If the class isn't to support days, and it also doesn't have and accessor for it, this should actually fail (or, remove the checks and support it)

src/eckit/types/Time.cc
```diff
@@ -27,80 +30,98 @@ inline void printTime(std::ostream& s, long n) {
     s << n;
 }
 
-Time::Time(long seconds) :
+Time::Time(long seconds, bool extended) :
     seconds_(seconds) {
```
pmaciel, line 34: This requires a seconds_(static_cast<double>(seconds))

src/eckit/types/Time.cc
pmaciel: Also, this printing of time as string is repeated in multiple places -- ideally, these places should be replaced with a operator std::string (that you already have)

pmaciel (changes requested):
The changes here are all an improvement 😄 👍 

I suggest in the future (because now isn't the right time) to move this class downstream, as it is not at the righ place on the pgen/mars-client stack (to my knowledge); it isn't "down enough", it is part of, and contains, MARS-specific interpretation of time, and should exist at the language parsing level instead.

tests/types/test_time.cc
```diff
@@ -0,0 +1,175 @@
...
+CASE("Time with unit (__h__m__s)") {
+    EXPECT(Time(2,0,0) == Time("2h"));
+    EXPECT(Time(2,0,0) == Time("0002H"));
+    EXPECT(Time(2,0,0) == Time("120m"));
+    EXPECT(Time(2,0,0) == Time("7200s"));
+
+    EXPECT(Time(0,3,0) == Time("3M"));
+    EXPECT(Time(0,3,0) == Time("180s"));
+
+    EXPECT(Time(1,23,45) == Time("1h23m45s"));
```
simondsmart, line 152: We ought to have tests that cover such things as 1h2m3s (I.e. where mins and secs are single digit, including following previous terms).

src/eckit/types/Time.cc
```diff
@@ -127,9 +147,9 @@ Time& Time::operator=(const Time& other) {
     return *this;
 }
 
-Time::Time(long hh, long mm, long ss) :
+Time::Time(long hh, long mm, long ss, bool extended) :
     seconds_(hh * 3600 + mm * 60 + ss) {
-    if (hh >= 24 || mm >= 60 || ss >= 60 || hh < 0 || mm < 0 || ss < 0) {
+    if (mm >= 60 || ss >= 60 || (!extended && (hh >= 24 || hh < 0 || mm < 0 || ss < 0))) {
```
simondsmart, line 152: IS there a reason we can't don't support 61 minutes, or 61 seconds, and lroll them into larger time units, given that we do that with hours here?

simondsmart (approved):
Looks good. Noting Pedro's comments as well. Please lightly tidy up, then merge.

PR #87, opened 2023-11-21
Increase ChannelBuffer size from 1024 to 4096

geier1993:
We are sometimes experiencing SEGFAULTS due to buffer overflows.

@raguridan may add more information.


dsarmany (approved):

dsarmany:
https://github.com/ecmwf/eckit/issues/89 has been raised to fix this correctly, which most likely will mean blocking logging if the buffer size is reached.

PR #90, opened 2023-11-22
Correct macOS specific downstream issues ECKIT-620

marcosbento:
This set of changes resolves build issues specific to Apple Clang 14 (on macOS 13).

Apparently this compiler doesn't work well with inner unnamed namespaces, and doesn't seem to generate the intended linkage symbols. This affected the definition of struct Title, which is a small internal utility to determine the description of an option based on the option type.

The solution was simply to move the Title class to a named namespace.

### DIFF
--- a/src/eckit/option/FactoryOption.cc
+++ b/src/eckit/option/FactoryOption.cc
@@ -12,6 +12,8 @@
 /// @author Tiago Quintino
 /// @date Apr 2015
 
+#pragma once
+
 #include <iostream>
 
 #include "eckit/config/Configuration.h"
--- a/src/eckit/option/FactoryOption.cc
+++ b/src/eckit/option/FactoryOption.cc
@@ -16,15 +16,11 @@
 
 #include <iostream>
 
-#include "eckit/config/Configuration.h"
 #include "eckit/config/Configured.h"
-
 #include "eckit/exception/Exceptions.h"
 #include "eckit/option/FactoryOption.h"
-#include "eckit/utils/Translator.h"
 
-namespace eckit {
-namespace option {
+namespace eckit::option {
 
 
 template <class T>
@@ -62,11 +58,9 @@ void FactoryOption<T>::copy(const Configuration& from, Configured& to) const {
 
 template <class T>
 void FactoryOption<T>::print(std::ostream& out) const {
-    out << "   --" << name_ << "=name"
-        << " (" << description_ << ")";
+    out << "   --" << name_ << "=name" << " (" << description_ << ")";
     out << std::endl << "     Values are: ";
     T::list(out);
 }
 
-}  // namespace option
-}  // namespace eckit
+}  // namespace eckit::option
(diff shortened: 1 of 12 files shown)


pmaciel:
I've taken the chance to further fix the build (re. templates + ecbuild), cleanup the include hierarchy and make the proprocessor directives consistent.

This was already a good cleanup, I just added a few sprinkles :-)

src/eckit/option/FactoryOption.cc
wdeconinck, line 1: I notice this file with `.cc` ext gets #included in `FactoryOption.h`
This is the wrong ext and compiler will try to compile this file directly.
Instead the ext could be `tcc` (for template cc)
See e.g. https://github.com/ecmwf/ecbuild/blob/develop/cmake/ecbuild_separate_sources.cmake#L35C2-L35C2
Where this file then also gets recognised as header file.

wdeconinck (changes requested):

src/eckit/option/FactoryOption.cc
pmaciel: Yes, that's true for FactoryOption.cc, VectorOption.cc and SimpleOption.cc (and likely, many other places in the code)

Then, these should also be removed from the respective (src/eckit/optios/CMakeLists.txt ecbuild_add_library) SOURCES and only present in the TEMPLATES part.

wdeconinck: Or as mentioned, tcc ext will also automatically work. As you choose.

wdeconinck (approved):

src/eckit/option/FactoryOption.cc
marcosbento: Looking at the `ecbuild_add_library` implementation, we have 2 options:
(a) #include .cc files, and tens list them as SOURCES and TEMPLATES when calling ecbuild_add_library
(b) have .tcc sources included as #include .tcc files, and then list them only as SOURCES , and let `ecbuild_add_library` automatically sort these template files out.

marcosbento: eckit seems to be following approach (a). I don't find any .tcc file in the code base.

I'm happy to align with approach (b) and make all necessary eckit updated in the future, but I not sure we should use this PR to make those changes.

wdeconinck:
Failure in pgen, but not a bug in eckit:
```
In file included from ../src/tools/pgen-compare.cc:17:
  ../src/pgen/compare/CompareProducer.h:88:18: error: ‘set’ in namespace ‘std’ does not name a template type
     88 |     mutable std::set<std::string> shown_;
        |                  ^~~
  ../src/pgen/compare/CompareProducer.h:24:1: note: ‘std::set’ is defined in header ‘<set>’; did you forget to ‘#include <set>’?
  ```

src/eckit/option/FactoryOption.cc
wdeconinck: Yes better to stay consistent with eckit style

wdeconinck:
I created PR in pgen @danovaro to address above issue.

Issue #89, opened 2023-11-22
Segmentation fault caused by buffer overflow in ChannelBuffer


dsarmany:
### What happened?

On LUMI, with huge amount of logging thanks to `MULTIO_DEBUG=1`, and the fact the filesystem there is sometimes slow, the default buffer size of 1024 may overflow and cause a crash.

### What are the steps to reproduce the bug?

Run an HPC experiment on LUMI, using RAPS and IFS+FESOMv2 or IFS+NEMOv4. Set `MULTIO_DEBUG=1`. There is a good chance (but no certainly) that the error will occur.

We have not been able to reproduce it in other contexts, but if the buffer-size is set to a small value, the expectation is that it should be reproducible more readily.

### Version

1.24.4

### Platform (OS and architecture)

LUMI/23.03 & cpeCray/23.03

### Organisation

ECMWF

simondsmart:
I want to clarify this further. We are seeing the symptom on LUMI. But the bug is always present.

If, for whatever reason, the flushing of the eckit::Log internal buffers stalls, it is possible to trigger a segfault by doing logging output. This has nothing to do with MULTIO, MULTIO_DEBUG, or LUMI.

This can (presumably) be simply reproduced by removing the flushing code from the eckit::Log infrastructure.

The thing that is nasty about this is that it would produce segfaults that are extremely difficult to reproduce, as it is likely that the user did not know that there had been a (potentially intermittent) performance issue on the logging filesystem. We need to make sure that eckit::Log cannot overflow.

geier1993:
Reproduction on LUMI (dev-g):

Reducing the ChannelBuffer size to 1 (or another low value) & use multio-hammer with a lot of threads:
`MULTIO_SERVER_CONFIG_PATH=$CLIMATEDT_RUNS_HOME/ifs-bundle/ifs-bundle/source/multio/tests/multio/config FDB_DEBUG=0 MULTIO_DEBUG=0 gdb --args ./bin/multio-hammer --transport=thread --nbclients=16 --nbservers=8`

Compiling with BIT or DEBUG (shows line numbers...) - both consistently point to `eckit::WrapperTarget::write`:

```
#0  0x00001555529810a6 in eckit::WrapperTarget::write (this=0x293690,
    start=0x3e1000 <error: Cannot access memory at address 0x3e1000>, end=0x36c11f "")
[...]

PR #95, opened 2023-11-23
Add defensive pointer boundary checks to log targets to avoid SEGFAUL…

geier1993:
…TS with raceconditions #89

### DIFF
--- a/src/eckit/log/ChannelBuffer.cc
+++ b/src/eckit/log/ChannelBuffer.cc
@@ -23,11 +23,9 @@ namespace eckit {
 
 //----------------------------------------------------------------------------------------------------------------------
 
-ChannelBuffer::ChannelBuffer(std::size_t size) :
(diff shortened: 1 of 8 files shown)


geier1993:
We are still discussing a way to make the ChannelBuffer threadsafe (or maybe to find another solution)

dsarmany (approved):

src/eckit/log/ChannelBuffer.cc
```diff
@@ -69,9 +67,10 @@ void ChannelBuffer::reset() {
 
 bool ChannelBuffer::dumpBuffer() {
     if (target_) {
-        target_->write(pbase(), pptr());
+        // Explicitly check that `pptr()` is not larger than end of buffer. Racecondition can end up adding larger values.
```
simondsmart, line 70: Can we be explicit that this is a race condition in the std library, rather than here.

PR #99, opened 2024-01-09
Feature/qhull

pmaciel:
This PR adds a ConvexHull base class and an implementation ConvexHullN which is templated to the number of dimensions of calculating a convex hull on. This exists in the eckit::maths library as discussed, and it is disabled by default (enable with cmake -DENABLE_CONVEX_HULL=ON ...)

The convex hull calculation itself is implemented by Qhull, hidden behind the interfacing classes, but exposing enough interface to catch exceptions including those internal to Qhull, so that the returned message(s) is meaningful to client code. Tests (and minimal examples) are in test_convex_hull.cc.


tlmquintino:
I've swapped @simondsmart for @danovaro as reviewer given his experience with computational geometry.

src/eckit/maths/Qhull.h
```diff
@@ -0,0 +1,57 @@
...
+#pragma once
+
+
+#include <string>
+#include <vector>
+
+#include "eckit/exception/Exceptions.h"
+#include "eckit/maths/ConvexHull.h"
+
+#include "libqhullcpp/Qhull.h"
```
wdeconinck, line 21: I would avoid this "libqhullcpp/Qhull.h" include in this place, so that qhull can stay a private dependency of eckit.

Following approach is what I used in atlas:
```
class Qhull {
...
private:
  // forward declaration
  struct Qhull_;
  std::unique_ptr<Qhull_> qh_;
};
```

and in the `.cc` file:
```
#include "libqhullcpp/Qhull.h"

struct Qhull::Qhull_ : public orgQhull::Qhull {
    using orgQhull::Qhull::Qhull;
};

// Qhull constructor
Qhull::Qhull(.....) {
    qh_ = std::make_unique<Qhull>(......);
    ...
}

File: src/eckit/maths/Qhull.h [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+
+    // -- Constructors
+
+    Qhull(size_t N, const coord_t& coord, const std::string& command);
+
+    // -- Overridden methods
+
+    std::vector<size_t> list_vertices() const;
+    std::vector<std::vector<size_t>> list_facets() const;
+    std::vector<Triangle> list_triangles() const;
```
wdeconinck, line 48: list_triangles() only works in particular case I suppose. This should be documented.

src/eckit/maths/ConvexHullN.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+    using coord_t = typename Qhull::coord_t;
+
+    explicit ConvexHullN(const coord_t& coord) : qhull_(N, coord, "QJ") {}
+    explicit ConvexHullN(const std::vector<std::vector<double>>& coord_v) : ConvexHullN(convert_vector_v(coord_v)) {}
+    explicit ConvexHullN(const std::vector<std::array<double, N>>& coord_a) : ConvexHullN(convert_vector_a(coord_a)) {}
+
+    std::vector<size_t> list_vertices() const override { return qhull_.list_vertices(); }
+    std::vector<std::vector<size_t>> list_facets() const override { return qhull_.list_facets(); }
+    std::vector<Triangle> list_triangles() const override { return qhull_.list_triangles(); }
```
wdeconinck, line 34: list_triangles() only works in particular case. This should be documented.

src/eckit/maths/CMakeLists.txt
```diff
@@ -1,16 +1,24 @@
...
+    Eigen.h
+    Lapack.h
+    Lapack.cc
+    Matrix.h
+    MatrixEigen.h
+    MatrixLapack.h)
+
+if(eckit_HAVE_CONVEX_HULL)
+    list(APPEND eckit_maths_sources ConvexHull.h ConvexHullN.h Qhull.cc Qhull.h)
+    list(APPEND eckit_maths_public_libs Qhull::Qhull)
```
wdeconinck, line 12: With previous comment about private includes within code, this could become private instead of public

src/eckit/maths/ConvexHull.h
```diff
@@ -0,0 +1,43 @@
...
+    ConvexHull(ConvexHull&&)      = delete;
+
+    virtual ~ConvexHull() = default;
+
+    void operator=(const ConvexHull&) = delete;
+    void operator=(ConvexHull&&)      = delete;
+
+    virtual std::vector<size_t> list_vertices() const            = 0;
+    virtual std::vector<std::vector<size_t>> list_facets() const = 0;
+    virtual std::vector<Triangle> list_triangles() const         = 0;
```
wdeconinck, line 36: list_triangles() only works in particular case I suppose. This should be documented.

wdeconinck (commented):
See below comments.
I'm wondering if the list_triangles() function should be there at all as it is a special case. Open to discuss.
Atlas already has a bespoke Qhull triangulation API in any case.

tests/maths/test_convex_hull.cc
```diff
@@ -0,0 +1,240 @@
...
+        };
+
+        for (const auto& test : tests) {
+            try {
+                Qhull(test.N, test.coord, test.command);
+                EXPECT(false);
+            }
+            catch (const Qhull::Exception& e) {
+                EXPECT_EQUAL(test.errorCode, e.errorCode);
+                EXPECT_EQUAL(test.what, std::string(e.what(), test.what.length()));
```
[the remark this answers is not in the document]
pmaciel: I don't think they are necessarily stable, but I do think they don't change often (and we'd possibly like to be sure if/when they do). I didn't try to be comprehensive, but put some cases where convex hull has problems with. This interface (of exception handling) and test was part of the requirements, that any errors are "trappable" and a human-readable message relayed to client code

src/eckit/maths/Qhull.h
pmaciel: Yes I wasn't too happy with this, I will adapt a similar solution

src/eckit/maths/Qhull.h
pmaciel: Yes, but triangulation is rather the common usage in most cases of both convex hull and delaunay triangulations, that's why I've included it. The Triangle element is a "flat" as can be. I didn't go as far as to specify a Polygon element though, I think it's overkill (and there is intersection with the ::geo/::geometry parts too), but it would be correct though.

In random lists of points (in most geometry cases for our applications, but not necessarily the case for polytope (where many of the axes are aligned or of the same size), the result is a triangulation. The common examples using Qhull (that I found) are with "Qt" which enforces this, so,  I think it is natural to have it as base functionality

src/eckit/maths/ConvexHullN.h
pmaciel: (as above)

src/eckit/maths/CMakeLists.txt
pmaciel: Yes I'm aware, will fix this

src/eckit/maths/Qhull.h
pmaciel: I've made it private with [commit 1c0c4073ae](https://github.com/ecmwf/eckit/pull/99/commits/1c0c4073ae96c46d173413a63db77ed79caa43f0). I've opted for a naked pointer instead (of course with proper new/delete)

pmaciel:
> See below comments. I'm wondering if the list_triangles() function should be there at all as it is a special case. Open to discuss. Atlas already has a bespoke Qhull triangulation API in any case.

I'm not sure of the triangulation, but it is rather the common than the special case (hence why I included it). It was in fact the more "testable" situation as in that simpler geometry checks are needed to check correctness.

src/eckit/maths/Qhull.h
wdeconinck: Fantastic!

wdeconinck:
> > See below comments. I'm wondering if the list_triangles() function should be there at all as it is a special case. Open to discuss. Atlas already has a bespoke Qhull triangulation API in any case.
> 
> I'm not sure of the triangulation, but it is rather the common than the special case (hence why I included it). It was in fact the more "testable" situation as in that simpler geometry checks are needed to check correctness.

What I refer to is that currently `list_facets()` also returns the triangles, albeit as `vector<vector<size_t>>` instead of `vector<array<size_t,3>>`

I would suggest an additional API for `list_facets()` to avoid the memory fragmentation and the counts of dynamic allocations / deallocations of `vector<vector>>`. An approach that returns two or three `vector`:
- two vectors:
  * vector 1:  contiguous array that contains all with vertices of the facets
  * vector 2: array with offsets where a facet starts in the above array, size == nb_facets
 - three vectors:
   * the two vectors from above, AND a third vector (size == nb_facets) that shows the number of vertices for each facet

The third vector is for the case that you'd want some byte alignment where a new facet starts in vector 1, and the nb_facets cannot be inferred from the difference in offsets in vector 2.

An additional API could wrap these 2/3 vectors  in a `Facets` return type with an inline accessor for a `Facet` from these vectors

pmaciel:
Following the recommendation, I decided to follow for the 2-vector solution while removing the _list_triangles_ in favour of a overloaded _list_facets(size_t n)_. For simplicity and to make the return type compatible in the two list_facets methods, the argument _n_ filters the factes for those of the expected size.

This is keeping in mind the considered case (of Polytope?). I agree that for mesh generation one could make the API even more specific, but in that case I would instead pass a single vector, addressable as _n*i + j_ as a block connectivity (as we've used before).

@danovaro @jameshawkes @tlmquintino @wdeconinck what do you @tithink?

wdeconinck:
Thanks for addressing my recommendations. 
For my understanding, Is the intention of facets_n map like this?
```
for( auto& [nb_vertices, nb_facets] : qhull.facets_n() ) {
  auto facets_with_nb_vertices = qhull.facets(nb_vertices);
  ASSERT( facets_with_nb_vertices.size() == nb_vertices * nb_facets );
}
```
So that for each type of facets you get a separate list?

wdeconinck:
An additional comment to @jameshawkes , I understood from @tlmquintino that you are currently using CGAL in C++ somewhere, but I can only see python code for polytope. Could you show the use case for this, as this is basically supposed to be used for polytope.

wdeconinck:
Amazing @pmaciel! Thanks for extending the square pyramid test! This is good to merge a.f.a.i.c.

jameshawkes:
> An additional comment to @jameshawkes , I understood from @tlmquintino that you are currently using CGAL in C++ somewhere, but I can only see python code for polytope. Could you show the use case for this, as this is basically supposed to be used for polytope.

We don't currently use CGAL, but as we are building a port of polytope to C++ (currently in Python) we need a convex hull interface we can use. QHull is what we use in Python, via scipy.spatial.

You can see how we currently use it in Python here:
https://github.com/ecmwf/polytope/blob/d4746f36cb70fda7ae6c58d8d24112c48b9056c0/polytope/engine/hullslicer.py#L173C5-L187C56

The interface for passing args in and receiving a result is pretty standard. The main challenge is the error handling. You can see in that code we are switching based on the error description to detect some failure edge cases which are important to us (detecting if the input points are "flat" or if the input shape is lower dimension than the space it occupies). This interface through scipy isn't great.

I discussed a bit with @pmaciel about the need to handle this error-handling a bit better in our C++ wrapper, if possible.

wdeconinck:
The exception handling is now exceptional! 😅

I think no-one else is probably interested in reviewing this but I'd give some time to the end of the week.
If nobody else reviewed by then, I (or you) can merge this.

PR #104, opened 2024-01-24
Make construction and access to of Log::debug<Lib> safe in multiple t…

simondsmart:
…hreads

### DIFF
--- a/src/eckit/system/Library.cc
+++ b/src/eckit/system/Library.cc
@@ -36,8 +36,7 @@ namespace eckit::system {
 
 //----------------------------------------------------------------------------------------------------------------------
 
-Library::Library(const std::string& name) :
-    name_(name), prefix_(name), debug_(false) {
+Library::Library(const std::string& name) : name_(name), prefix_(name), debug_(false) {
 
     LibraryManager::enregister(name, this);
 
@@ -120,19 +119,10 @@ Channel& Library::debugChannel() const {
         return *it->second;
     }
 
-    std::string s = prefix_ + "_DEBUG";
-
-    std::unique_ptr<Channel> newChannel;
-    if (debug_) {
-        newChannel = std::make_unique<Channel>(new PrefixTarget(s));
-    }
-    else {
-        newChannel = std::make_unique<Channel>();
-    }
-
-    Channel& retval(*newChannel);
-    debugChannels.emplace(this, newChannel.release());
-    return retval;
+    return *debugChannels
+                .emplace(this, debug_ ? std::make_unique<Channel>(new PrefixTarget(prefix_ + "_DEBUG"))  //
+                                      : std::make_unique<Channel>())                                     //
+                .first->second.get();
 }
 
 const Configuration& Library::configuration() const {
@@ -149,7 +139,8 @@ const Configuration& Library::configuration() const {
 
     Log::debug() << "Parsing Lib " << name_ << " config file " << cfgpath << std::endl;
 
-    eckit::Configuration* cfg = cfgpath.exists() ? new eckit::YAMLConfiguration(cfgpath) : new eckit::YAMLConfiguration(std::string(""));
+    eckit::Configuration* cfg
+        = cfgpath.exists() ? new eckit::YAMLConfiguration(cfgpath) : new eckit::YAMLConfiguration(std::string(""));
 
     Log::debug() << "Lib " << name_ << " configuration: " << *cfg << std::endl;
 
--- a/src/eckit/system/Library.cc
+++ b/src/eckit/system/Library.cc
@@ -30,6 +30,7 @@
 #include "eckit/system/SystemInfo.h"
 #include "eckit/thread/AutoLock.h"
 #include "eckit/thread/Mutex.h"
+#include "eckit/thread/ThreadSingleton.h"
 #include "eckit/utils/Translator.h"
 
 namespace eckit::system {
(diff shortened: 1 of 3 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=104) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=104) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/eckit?pullRequest=104) it.</sub>

src/eckit/system/Library.cc
```diff
@@ -109,23 +109,30 @@ std::string Library::libraryPath() const {
...
+        newChannel = std::make_unique<Channel>(new PrefixTarget(s));
     }
     else {
-        debugChannel_.reset(new Channel());
+        newChannel = std::make_unique<Channel>();
     }
 
-    return *debugChannel_;
+    Channel& retval(*newChannel);
+    debugChannels.emplace(this, newChannel.release());
```
geier1993, line 134: I'd try to avoid using `release()` on unique pointer.
We can directly `emplace` and use the result iterator:

```
return *debugChannels.emplace(this, 
    debug_ ? std::make_unique<Channel>() 
                   : std::make_unique<Channel>(new Prefix Target(s))
).first->get();
```

If you don't like the conditional inside the emplace, I'd put it in a function/lambda. But all in all I think it's better to emplace directly instead of releasing a unique pointer while referring to its value.

Technically also instead of `release()` a `std::move(newChannel)` should also keep the pointer inside alive - but avoiding it at all might be cleaner.

geier1993 (approved):
Just a question for completeness/transparency:
When we were handling the race condition inside the targets, we decided not to avoid the race condition (through a threadlocal channel) and make it harmless instead.
It was said that we still want to have the performance and do not care about dirty logs. 
Is there a reason why this decision has changed?

geier1993:
Multio has failing CI runs due SEGFAULTS we could trace back to this race condition (although it's not happening very often).
However, as multio is downstream for a lot of packages it would be preferable to fix the root throug this PR instead of removing the test.

I will address the changes in my comment and then hope to be able to merge this branch

src/eckit/system/Library.cc
geier1993: I'm making changes now  and hope that we can merge this soon

geier1993:
Okay, we can some new SEGFAULTS in multio now. That's bad ... need to investigate. Downstream ci helped in this case

geier1993 (changes requested):
New segfaults in multio tests are occuring now. others are fixed

Update:
FDB performs a lot of debug logs in the Destructors of other static objects... in combination with thread_local this causes SEGFAULTS on library shutdown as destruction order is not clear. 
Hard to get around. Trying a different approach with a map of thread ids and frequent clean ups......

Ideally we need a RW lock...  using std::shared_mutex is not compiling on Mac currently (?! linking never ends here... just read about some trouble online)

simondsmart:
This feels like it is ... overly complicated. And not totally clear that it will solve the issue (as opposed to likely solves the issue).

If the segfaults are constrained to the destructors of static objects can we not:

i) Enforce that any debug statements in those destructors use Log::debug() rather than Log::debug<LIB>() -- i.e. use the global object. This should make the construction/destruction order more straightforward to reason

ii) alternatively, in the same way that we have "if (!Main::ready())" in the debug() function, we could add a special post-main flag that gets called (presumably by the destructor of one one of these ThreadSingleton or Channel objects) which we use to  activate a different output pathway.

geier1993:
> This feels like it is ... overly complicated. And not totally clear that it will solve the issue (as opposed to likely solves the issue).
> 

Compared with the initial solution with thread local map with pointers to static library object, this approach is reversed and stores channels directly in the static library object and maps thread ids - up to this point I would not consider it as more complex.
However making sure that exiting threads remove their entry in the maps is where dependency & life time problem of static / thread_local objects hook in... but to resolve these kind of problems, the safest option will always be making use of shared obects both statics can access safely.


> If the segfaults are constrained to the destructors of static objects can we not:
> 
> i) Enforce that any debug statements in those destructors use Log::debug() rather than Log::debug() -- i.e. use the global object. This should make the construction/destruction order more straightforward to reason

This is a lot of effort and very error prone. It involves a lot of changes downstream, it's likely to forget something and we have no usage to check/enforce this "policy". IMO this will result in recurring occurrences of this phenomenon.


> ii) alternatively, in the same way that we have "if (!Main::ready())" in the debug() function, we could add a special post-main flag that gets called (presumably by the destructor of one one of these ThreadSingleton or Channel objects) which we use to activate a different output pathway.

Also in this case, we would need to to "sync" between two static/thread_local obects of different lifetime. To do so we eventually need to make use of a shared_ptr again....

However, a valid option to bring in a flag may be to use a share_ptr and a weak_ptr on the other side ... locking the weak_ptr would be similar to a check ?

geier1993 (approved):
bless the ThreadSingleton

mcakircali:
This causes crash when Log::debug() called in a class dtor
PANIC: ::pthread_setspecific(key_, value) in  (src/eckit/thread/ThreadSingleton.h +97 instance)

PR #105, opened 2024-01-24
Deprecate access to `const Value& Configuration::get()`

wdeconinck:
`const eckit::Value& eckit::Configuration::get()` is "public" and likely abused in various places, leaking the `eckit::Value` abstraction.

This pull request marks this function as deprecated, and warns for it to be removed in a next release.

Further changes to `LocalConfiguration` have been made so that classes like e.g. `eckit::codec::MetaData` or `atlas::Config` which derive from `LocalConfiguration` don't need internal access to `eckit::Value`.

I have modified the `atlas::Config` class respectively and can commit that once this PR is merged.

This relates to PR #101 as without access to `eckit::Value` one will no longer be able to use `eckit::Value` for introspection.

### DIFF
--- a/src/eckit/config/LocalConfiguration.h
+++ b/src/eckit/config/LocalConfiguration.h
@@ -15,7 +15,9 @@
 #ifndef eckit_LocalConfiguration_H
 #define eckit_LocalConfiguration_H
 
+#include <initializer_list>
 #include <vector>
+#include <type_traits>
 
 #include "eckit/config/Configuration.h"
 #include "eckit/config/Configured.h"
@@ -42,6 +44,8 @@ public:  // methods
 
     ~LocalConfiguration() override;
 
+    LocalConfiguration& set(const Configuration& other);
+
     LocalConfiguration& set(const std::string& name, const std::string& value) override;
     LocalConfiguration& set(const std::string& name, const char* value) override;
     LocalConfiguration& set(const std::string& name, bool value) override;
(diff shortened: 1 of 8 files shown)


src/eckit/config/LocalConfiguration.h
```diff
@@ -76,6 +77,18 @@ class LocalConfiguration : public Configuration, public Configured {
...
+    template <typename ConfigurationT,
+              typename = std::enable_if_t<std::is_base_of_v<Configuration, std::decay_t<ConfigurationT>>>>
+    LocalConfiguration& set(const std::string& name, std::initializer_list<ConfigurationT>&& value) {
+        std::vector<const Configuration*> abstract_pointers;
+        abstract_pointers.reserve(value.size());
+        for (auto& v: value) {
+            abstract_pointers.emplace_back(&v);
+        }
+        return set(name, abstract_pointers.data(), abstract_pointers.size());
+    }
```
wdeconinck, line 89: This change was required to support a failing fdb5 test.
e.g. https://github.com/ecmwf/eckit/actions/runs/7641531854/job/20819349927#step:3:2397

pmaciel (approved):
I certainly appreciate the adherence to the architecture principles and the cleanup -- and the testing! 5 ⭐

geier1993:
Currently we use the `eckit::Value` to get values for a key and check their type via a long list of function calls `isMap`, `isList`, `isNumber`, ...
With the other PR on introspective features, this gets slightly worse. The long list then always accepts a key and will perform a lot of unnecessary lookups.

If you really don't want to expose a value, it would be at least helpful to have a visitor gets called with the right type.
We could implement the visitor for all Contents, Value and the LocalConfiguration.

The other option would be to expose an type enum.

tlmquintino:
Yes I think this is a correct step. I know it makes things harder for the what @geier1993 mentioned, but we need to address that differently.

tlmquintino (approved):
good step forward and cleanup.

PR #110, opened 2024-03-11
Fix warnings

wdeconinck:
Fix warnings with Intel compiler.
Many are also in common with other compilers (e.g. related to partially overridden method).
Some are related to unrecognised clang pragmas with Intel compiler.

Getting compilation of eckit without all the warnings should be beneficial with Intel compiler as this is our main targeted compiler at ECMWF.

For reviewers: It would be best to look at the changes on a commit-per-commit basis.

### DIFF
--- a/src/eckit/mpi/Group.cc
+++ b/src/eckit/mpi/Group.cc
@@ -32,7 +32,7 @@ public:
 
     virtual NullGroupContent* union_group(const GroupContent&) const { return new NullGroupContent(); }
 
-    virtual size_t size() const { return -1; }
+    virtual size_t size() const { return 0; }
 
     virtual int rank() const { return -1; }
 
--- a/src/eckit/web/HttpResource.cc
+++ b/src/eckit/web/HttpResource.cc
@@ -229,13 +229,10 @@ void HttpResource::dispatch(eckit::Stream&, std::istream&, HttpStream& out, Url&
 
                 std::ostringstream oss;
                 oss << "Unsupported HTTP method " << method << " url=" << url;
-                throw eckit::MethodNotYetImplemented(oss.str());
+                throw eckit::NotImplemented(oss.str());
             }
             catch (eckit::HttpError& e) {
                 error(url, out, e, e.status());
-            }
-            catch (eckit::MethodNotYetImplemented& e) {
-                error(url, out, e, HttpError::NOT_IMPLEMENTED);
             }
             catch (eckit::NotImplemented& e) {
                 error(url, out, e, HttpError::NOT_IMPLEMENTED);
--- a/tests/option/eckit_test_option_cmdargs.cc
+++ b/tests/option/eckit_test_option_cmdargs.cc
@@ -1015,6 +1015,11 @@ CASE("test_eckit_option_cmdargs_value_with_equals") {
 #ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#elif defined(__INTEL_COMPILER)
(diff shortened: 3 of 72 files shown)


tests/option/eckit_test_option_cmdargs.cc
```diff
@@ -1015,6 +1015,9 @@ CASE("test_eckit_option_cmdargs_value_with_equals") {
 #ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#elif defined(__INTEL_COMPILER)
+#pragma warning ( push )
+#pragma warning ( disable:1786 )
```
pmaciel, line 1020: Short description on this warning

src/eckit/mpi/Group.cc
```diff
@@ -32,7 +32,7 @@ class NullGroupContent : public GroupContent {
 
     virtual NullGroupContent* union_group(const GroupContent&) const { return new NullGroupContent(); }
 
-    virtual size_t size() const { return -1; }
+    virtual size_t size() const { return 0; }
```
pmaciel, line 35: Whoa good catch -- I'm surprised a big fish like this stood for so long!

src/eckit/web/HttpResource.cc
```diff
@@ -229,14 +229,11 @@ void HttpResource::dispatch(eckit::Stream&, std::istream&, HttpStream& out, Url&
 
                 std::ostringstream oss;
                 oss << "Unsupported HTTP method " << method << " url=" << url;
-                throw eckit::MethodNotYetImplemented(oss.str());
+                throw eckit::NotImplemented(oss.str());
```
pmaciel, line 232: We should deprecate the replaced exception (MethodNotYetImplemented) as there is no code in eckit supporting this exception anymore and none of these projects: atlas, atlas-orca, bufrdc_tables, dhskit, eccodes, eckit, fckit, fdb, fdb5, mars-adsm, mars-client, mars-hpss, mars-server, metkit, mir, multio, odc, pgen, transi, ...

pmaciel (commented):
Very minor changes

src/eckit/mpi/Group.cc
wdeconinck: It's in NullGroupContent which has other mechanisms to check that the group is not valid. Probably one never called this function.

tests/option/eckit_test_option_cmdargs.cc
wdeconinck: The description is on line 1014 just above. It is hidden by this diff.

src/eckit/web/HttpResource.cc
wdeconinck: OK, I don't know the history of this one. If there is no occurrence anywhere in the stack we can just remove it rather than deprecate?
If no occurrence anymore, which projects would have used it in the past, and how long ago?

wdeconinck: Deprecated with commit [8510f4e](https://github.com/ecmwf/eckit/pull/110/commits/8510f4e9e66dd1f3bcdcd424738baf2dd7d07ea2)

pmaciel (approved):

wdeconinck:
@pmaciel I have added some more commits to avoid even more warnings.
In particular one you will like is [8d2fcd4](https://github.com/ecmwf/eckit/pull/110/commits/8d2fcd492530dfe4781aa95566e21da029f2ff62) which avoids warnings from Eigen headers, by using the SYSTEM keyword in `target_include_directories`

PR #111, opened 2024-03-18
Fix `normalise_angle` bit-identicality

wdeconinck:
Previously merged PR #103 caused non-bit identical results and causes downstream tests to fail.
This PR fixes this by reintroducing the older `normalise_angle` formula, only in the range where we were usually expecting the bit-identical results.


pmaciel (approved):
Yes; let's keep in mind to remove the threshold once we need to change the base comparison

pmaciel:
I've added this to further avoid normalise even if there's a 0*360 involved, with
            auto diff       = a - minimum;
            return 0. <= diff && diff < 360. ? a : modulo_360(diff) + minimum;

PR #112, opened 2024-03-21
Add Regex::escape function to escape special regex chars from a string

geier1993:
In addition to a bug fix with FDB5 we need a mechanism to properly escape regex char from a string.

FDB5 uses paths in regex without escaping special chars. This causes problems for paths with characters like + or *.

### DIFF
--- a/src/eckit/utils/Regex.cc
+++ b/src/eckit/utils/Regex.cc
@@ -114,6 +114,39 @@ Regex& Regex::operator=(const Regex& other) {
     return *this;
 }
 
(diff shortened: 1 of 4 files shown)


geier1993:
Will add more unit tests for the regex in general...

ChrisspyB:
Related to https://git.ecmwf.int/projects/MARS/repos/fdb5/pull-requests/93/overview

simondsmart (approved):
This looks good.

Issue #71, opened 2023-05-11
Nested SECTIONs in tests do not appear to work


fmahebert:
### What happened?

The test case `test_polygon` uses nested SECTION scopes, and it appears that lines of code in inner SECTION scopes are not run. I think `test_polygon` may be the only eckit test using this nesting.

### What are the steps to reproduce the bug?

1. In file `test_polygon.cc` below line 240 (inside nested SECTION "Contains edge"), add new line of code `EXPECT(false);`. Expect test `ctest -R eckit_test_geometry_polygon` to fail, but it passes.
2. To show expected behavior, move this new line of code to above line 240 (in outer SECTION "Simple rectangular polygon"). Expect test to fail, and it does.

### Version

v1.20.2

### Platform (OS and architecture)

macOS 12.6.4; intel i7

wdeconinck:
@pmaciel do you have a fix for this in another branch?

pmaciel:
> @pmaciel do you have a fix for this in another branch?

My fix for it was to flatten the hierarchy, ie remove the nested SECTIONs. The work I have isn't ready to be merged, so I suggest to flatten here as well?

PR #116, opened 2024-04-06
Feature/geo

pmaciel:
This PR provides minimal functionality for a grid library, under eckit::geo and resulting in libeckit_geo.dylib. It is meant as a (much) larger superset of the eckit::geometry library which it intends to replace, as well as a very large rewrite. 

It has basic support for the following:

- regular ll, regular gg, reduced gg, ORCA, HEALPix grids (ring & nested); They can
  - iterate (inside eg. a for loop)
  - generate a list of points (a pair of lat/lon vectors, or a single list of points)
  - generate a spec ("gridspec"), which can be reused to recreate the same grid
  - be instantiated from a spec ("gridspec")
- regular ll, regular gg, reduced gg grids can crop
- the framework for Projections, Figure (shape and size), is in place
- all new class hierarchies are fully adequate for Python interfaces (if need be), that is, the classes are components

Everything is done "as late as possible", and the expensive calculations are cached. There is sufficient support for ORCA grids cached to/from disk to support the existing stack.

What's not there:
- ordering (GRIB scanningMode) handling ~~(incl. HEALPix ordering: nested but that is simple to add now)~~
- projections (incl. rotations), there is heavy testing done already (so it is simple to add now, one-by-one)

This is meant as a starting point. If one aims for full complete integration with eccodes, I would say we are 75% of the way there -- but the last % take longer.

This considers eccodes + atlas + mir + multio (the ORCA and HEALPix parts) + earthkit-regrid. There is also a parallel effort that can trigger most of the above functionality from eccodes, which has been considered for this design so it is suitable, with an equivalent grib_get_data (and an additional grib_get_data_diff) already supporting integration into eccodes (the first project with eckit::geo upstream of eccodes).


wdeconinck:
Is it the intention to remove the feature CONVEX_HULL ?

pmaciel:
> Is it the intention to remove the feature CONVEX_HULL ?

No, I accidentaly removed it when splitting branches. Thanks for spotting

wdeconinck:
OK, thanks. There's still a change in how Qhull is detected after your last commit. Probably best to leave it as it was, without "REQUIRED_PACKAGES", and use `find_package(Qhull)` further. If a change is warranted, best in a separate PR.

CMakeLists.txt
```diff
@@ -195,7 +213,7 @@ endif()
 
 ecbuild_add_option( FEATURE CURL
                     DESCRIPTION "Curl library for transfering data with URLs"
-                    REQUIRED_PACKAGES "CURL 7.20" )
```
wdeconinck, line 198: This is a revert of commit 391e6f414d, why?
This triggers some deprecation warning..

CMakeLists.txt
```diff
@@ -264,6 +282,16 @@ ecbuild_add_option( FEATURE AIO
                     CONDITION ${AIO_FOUND}
                     DESCRIPTION "support for asynchronous IO")
 
+### PROJ support
+
+ecbuild_add_option( FEATURE PROJ
+                    DEFAULT OFF
+                    DESCRIPTION "support PROJ-based projections"
+                    REQUIRED_PACKAGES "PROJ VERSION 9.2" )
```
wdeconinck, line 290: Probably no need for "PROJ VERSION 9.2", just "PROJ 9.2", see CURL comment above.

pmaciel:
> OK, thanks. There's still a change in how Qhull is detected after your last commit. Probably best to leave it as it was, without "REQUIRED_PACKAGES", and use `find_package(Qhull)` further. If a change is warranted, best in a separate PR.

No it's fine, you're right. I had the right code, but I forked from a while back and did the same with qhull too -- so it was confusing. Now fixed

tests/geo/spec.cc
```diff
@@ -0,0 +1,381 @@
...
+             C()},
+        };
+
+        for (const auto& [user, gridspec] : tests) {
+            Log::info() << user << " -> " << gridspec << std::endl;
+
+            try {
+                std::unique_ptr<const Spec> spec(GridFactory::make_spec(user));
+                EXPECT(spec);
+
```
sandorkertesz, line 300: This code should probably test the contents of the generated `spec` not only its existence.

src/eckit/geo/range/RegularLatitude.h
```diff
@@ -0,0 +1,35 @@
...
+
+
+namespace eckit::geo::range {
+
+
+class RegularLatitude final : public Regular {
+public:
+    // -- Constructors
+
+    explicit RegularLatitude(double inc, double a, double b, double ref, double eps = 0.);
```
wdeconinck, line 25: What is the use of the ref argument?

src/eckit/geo/range/Regular.cc
```diff
@@ -0,0 +1,79 @@
...
+
+    a(__a);
+    b(__b);
+    resize(n);
+}
+
+
+Fraction Regular::increment() const {
+    ASSERT(1 < size());
+    return Fraction(std::abs(b() - a()) / static_cast<double>(periodic() ? size() : (size() - 1)));
```
wdeconinck, line 52: Why not use cheaper `Fraction{top, bottom}` constructor? https://github.com/ecmwf/eckit/blob/06e074cf23fbff95bad292e378d706f56329f276/src/eckit/types/Fraction.h#L44C5-L45C1

This is probably relatively expensive calculation? Could it be cached as data member?

src/eckit/geo/range/RegularLatitude.h
pmaciel: When you define a regional regular_ll grid with area=n/w/s/e, the points will include the point defined at w/s by default. The ref defines a different point to include (or +- increments until it is included in area)

This is to support "shifted" and "non-shifted" grids for the same definition of area and increments.

src/eckit/geo/range/Regular.cc
pmaciel: Yes it is expensive, but also not useful except for building a grid spec, so I didn't see an interest in caching. The Fraction{top, bottom} cannot be used because a() and b() return doubles.

tests/geo/spec.cc
pmaciel: Yes, that was a missing test until the generation of grid specs was at all possible, I will add this.

src/eckit/geo/Domain.h
```diff
@@ -0,0 +1,114 @@
...
+    double east() const { return east_; }
+
+    bool includesNorthPole() const;
+    bool includesSouthPole() const;
+    bool isPeriodicWestEast() const;
+
+    bool contains(double lat, double lon) const;
+    bool contains(const Domain&) const;
+    bool empty() const;
+    double area(double radius) const;
```
wdeconinck, line 59: What does this do?

src/eckit/geo/grid/unstructured/UnstructuredFromGrid.cc
```diff
@@ -0,0 +1,90 @@
...
+}
+
+
+Grid::iterator UnstructuredFromGrid::cend() const {
+    return iterator{new geo::iterator::Unstructured(*this, size())};
+}
+
+
+std::vector<Point> UnstructuredFromGrid::to_points() const {
+    std::vector<Point> p;
```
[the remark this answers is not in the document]
wdeconinck: Actually looks like a memory error as p is not resized -> `p.size() = 0` and then `p[i]` is used in the subsequent loop.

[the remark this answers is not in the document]
pmaciel: You're right, that's not well written, working on it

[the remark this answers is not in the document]
pmaciel: Fixed; note that this will be removed

src/eckit/geo/grid/unstructured/UnstructuredFromGrid.cc
```diff
@@ -63,11 +61,14 @@ Grid::iterator UnstructuredFromGrid::cend() const {
 
 
 std::vector<Point> UnstructuredFromGrid::to_points() const {
-    std::vector<Point> p;
+    std::vector<Point> points;
```
shahramn, line 64: Why not compute size once and then use it in the loop?

pmaciel: Because there is no default constructor for the variant type Point

src/eckit/geo/Domain.h
pmaciel: This class is supposed to be a drop in replacement for mir's usage of the coverage of the grid area, as limited by 2 parallels and 2 meridians, and with double as a coordinate type (as distinguished from the mir's BoundingBox). But, it is overlapping in concept with the eckit::geo::BoundingBox, and this will be updated with a more general base class "Area" (the discussion hasn't yet taken place).

So I'm removing this now to make it more clear

wdeconinck: I was wondering what `double area(radius)` is supposed to return. A circular area around some kind of centrepoint?

src/eckit/geo/Figure.h
```diff
@@ -0,0 +1,121 @@
...
+
+namespace eckit::geo {
+class Spec;
+}
+
+
+namespace eckit::geo {
+
+
+class Figure {
```
wdeconinck, line 29: It is not obvious to me what a Figure means in `geo` context. Perhaps a few lines of comments with documentation could be helpful

src/eckit/geo/Domain.h
```diff
@@ -0,0 +1,114 @@
...
+    double north() const { return north_; }
+    double west() const { return west_; }
+    double south() const { return south_; }
+    double east() const { return east_; }
+
+    bool includesNorthPole() const;
+    bool includesSouthPole() const;
+    bool isPeriodicWestEast() const;
+
+    bool contains(double lat, double lon) const;
```
wdeconinck:
```suggestion
    bool contains(double lon, double lat) const;
```
For consistency with PointLonLat, and we could add another overload for `bool contains(const PointLonLat&) const`

src/eckit/geo/Iterator.h
```diff
@@ -0,0 +1,120 @@
...
+
+#pragma once
+
+#include <type_traits>
+
+#include "eckit/geo/Point.h"
+
+
+namespace eckit::geo {
+class Grid;
```
wdeconinck, line 21: Unused forward declaration eckit::geo::Grid

src/eckit/geo/Point.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "eckit/geo/Point2.h"
+#include "eckit/geo/Point3.h"
+#include "eckit/geo/PointLonLat.h"
+
+
+namespace eckit::geo {
+
+
+using Point = std::variant<PointLonLat, Point2, Point3>;
```
wdeconinck, line 26: Note that `sizeof(Point) = 2 * sizeof(PointLonLat)`
(also an extra size_t for the internal variant index)
This means double! the memory if this were used in containers for "lon,lat".
Something that I'd have to be very careful about.

src/eckit/geo/grid/unstructured/UnstructuredFromGrid.cc
wdeconinck: You can add `std::monostate` as the very first alternative which is well behaved.
https://en.cppreference.com/w/cpp/utility/variant/monostate

src/eckit/geo/Point.h
pmaciel: This is by design -- Point is the generic argument for the Iterator and the Projection, so that we can explit polymorphism, chain projections and having the same interface for handling cartesian [X, Y, Z], cartesian [m], and geodetic coordinates without exploding the number of combinations in the interface. Note that the cost is only paid at construction (of each point), but it should be resuable as a temporary when used in an Iterator.

The usage in the Grid::to_points is another case wheer the penalty is fully paid, but that's intended for conversion to unstructured grid (which should not happen in or within performance-critical code, like a model) and interfacing with Python (or something else)

src/eckit/geo/Iterator.h
pmaciel: This is a convenience declaration to make inherited classed access in their declarations; Indeed I don't have to do it this way, but it is saving 3 (possibly 4 in the future) fwd declarations down the hierarchy

src/eckit/geo/Point.h
wdeconinck: Perhaps it is sufficient for `Grid::to_points` to return Point2 ?

src/eckit/geo/Domain.h
pmaciel: I've removed the class not to be a problem (it isn't really needed for eckit::geo, so it's a good removal!), but I have the same methods for eckit::geo::BoundingBox. There "double area(radius)" means the surface area represented by the bounding box on a sphere of given raidus. I see that it should instead be part of Figure (which accounts for that supporting geometry), jult like a great-circle distance and the like

src/eckit/geo/Point.h
pmaciel: This way of using a std::variant I can attach a projection to a "usual" iterator converting (lon,lat) to (x,y,z) (or, but I don't have an example, to do (x,y) to (x,y,z)) to build a k-d tree in a very compact way, and the number of temporaries is minimal

src/eckit/geo/Domain.h
pmaciel: Yes you're right, the interface for a Point or one of its variants is better (for the BoundingBox class, this one I deleted for clarity). This class should actually be named BoundBoxLonLat as most of it's logic is specific to that coordinate system, but it is currently this way due the widespread use

src/eckit/geo/range/RegularLatitude.h
wdeconinck: Could this be documented in code for future readings.

src/eckit/geo/grid/unstructured/UnstructuredFromGrid.cc
wdeconinck: Super you have added this, and fixed the memory problem.

src/eckit/geo/Point3.h
```diff
@@ -0,0 +1,102 @@
...
+    Point3& operator=(Point3&& other) {
+        container_type::operator=(other);
+        return *this;
+    }
+
+    // -- Members
+
+    double& X = container_type::operator[](0);
+    double& Y = container_type::operator[](1);
+    double& Z = container_type::operator[](2);
```
wdeconinck, line 65: Following code should not be able to compile:
```c++
void my_function( const Point3& p ) {
  p.X = 1;
  p.Y = 2;
  p.Z = 3;
}
```
meaning that I should not be allowed to change the internal data on a `const Point3&`.
I'm not entirely sure that this is ensured.

src/eckit/geo/Range.h
```diff
@@ -0,0 +1,71 @@
...
+
+    size_t size() const { return n_; }
+    double a() const { return a_; }
+    double b() const { return b_; }
+    double eps() const { return eps_; }
+
+    virtual bool periodic() const { return false; }
+
+    virtual Range* flip() const                             = 0;
+    virtual Range* crop(double crop_a, double crop_b) const = 0;
```
wdeconinck, line 47: I'd add the `[[nodiscard]]` attribute here, or return `std::unique_ptr<Range>` better.

src/eckit/geo/Projection.h
```diff
@@ -0,0 +1,122 @@
...
+    Projection& operator=(Projection&&)      = default;
+
+    // -- Methods
+
+    static std::string className() { return "projection"; }
+
+    virtual Point fwd(const Point&) const = 0;
+    virtual Point inv(const Point&) const = 0;
+
+    virtual Spec* spec() const = 0;
```
wdeconinck, line 68: Same comment: `[[nodiscard]]` or better return `std::unique_ptr<Spec>`.

src/eckit/geo/area/BoundingBox.h
```diff
@@ -0,0 +1,127 @@
...
+    bool isPeriodicWestEast() const;
+    bool containsNorthPole() const;
+    bool containsSouthPole() const;
+
+    bool intersects(BoundingBox&) const;
+
+    bool contains(double lat, double lon) const;
+    bool contains(const BoundingBox&) const;
+    bool empty() const;
+    double area(double radius) const;
```
wdeconinck, line 84: Could you add in comments the expected units for radius and returned area?

pmaciel:
@simondsmart poke!

src/eckit/maths/CMakeLists.txt
```diff
@@ -1,25 +1,26 @@
...
+    MatrixLapack.h
+)
 
-if(eckit_HAVE_CONVEX_HULL)
-    list(APPEND eckit_maths_sources ConvexHull.h ConvexHullN.h Qhull.cc Qhull.h)
-    list(APPEND eckit_maths_private_libs Qhull::Qhull)
+if( eckit_HAVE_ECKIT_GEO )
+    # for eckit::geo
+    list( APPEND eckit_maths_lib_srcs
+        Matrix3.h
```
simondsmart, line 13: Is this Matrix type really _only_ applicable with eckit geo?

simondsmart (commented):
This looks really nice overall, despite the number of small comments.
* You seem to really like things declared with __ at the start. This is supposed to be reserved, and I think all of these can just be put into an anonymous namespace
* Some of your types are returned with raw pointers. They either need [[nodiscard]], to return a std::unique_ptr, or to have a Value-like wrapper class that handles the pointer stuff internally.
* There are a couple of bits (commented) where I don't think the changes are geo-specific. And they could do with going elsewhere. Particularly in spec, I wonder if we could do well to revisit the Parametrisation stuff in general.

src/eckit/geo/util/arange.cc
```diff
@@ -0,0 +1,39 @@
...
+        std::vector<double> l(1, start);
+        return l;
+    }
+
+    const auto num = static_cast<size_t>((stop - start) / step) + 1;
+
+    std::vector<double> l(num);
+    std::generate_n(l.begin(), num, [start, step, n = 0ULL]() mutable {
+        return start + static_cast<double>(n++) * step;
+    });
```
wdeconinck, line 33: Very cool to learn about std::generate_n (thanks!) but probably overkill. We know `l` is just a vector and not some fancy container. Compilers will have easier time optimising following.
```c++
for( size_t n = 0; n < num; ++n ) {
  l[n] = start + static_cast<double>(n) * step;
}
```

src/eckit/geo/util/linspace.cc
```diff
@@ -0,0 +1,36 @@
...
+    if (num == 0) {
+        return {};
+    }
+
+    const auto step = num > 1 ? (stop - start) / static_cast<double>(endpoint ? (num - 1) : num) : 0;
+
+    std::vector<double> l(num);
+    std::generate_n(l.begin(), num, [start, step, n = 0ULL]() mutable {
+        return start + static_cast<double>(n++) * step;
+    });
```
wdeconinck, line 30: Same here as with `arange`: simple `for` loop would be preferred.

src/eckit/geo/util/mutex.h
```diff
@@ -0,0 +1,58 @@
...
+#endif
+
+
+namespace eckit::geo::util {
+
+
+#if defined(ECKIT_GEO_ECKIT_THREADS)
+
+
+using recursive_mutex = eckit::Mutex;
```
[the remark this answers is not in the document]
wdeconinck: Should we in fact not follw up with a refactoring eckit::Mutex et. al. to use std mutexes?
Then one could interchangebly use eckit::Mutex or std mutex, also in third party c++ libraries that would be linked in not relying on eckit.
At the same time we should also deeply investigate interoperability with OpenMP (e.g. locking with `#pragma omp critical`), which is the de facto threading model most HPC codes are using.

src/eckit/geo/util/reduced_classical_pl.cc
```diff
@@ -0,0 +1,1366 @@
...
+
+#include "eckit/exception/Exceptions.h"
+#include "eckit/geo/Cache.h"
+#include "eckit/geo/util.h"
+
+
+namespace eckit::geo::util {
+
+
+static const std::map<size_t, pl_type> CLASSICAL_PLS{
```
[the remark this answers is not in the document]
wdeconinck: I think this is analogous to atlas. There this is done to avoid every MPI rank to load the same data file upon loading a grid. This issue was also leading to eccodes' possibility to compile its definitions into a shared library.
For eckit::geo, I guess this could also become a configurable option.

src/eckit/geo/Cache.h
```diff
@@ -0,0 +1,101 @@
...
+#include "eckit/exception/Exceptions.h"
+#include "eckit/geo/util/mutex.h"
+
+
+namespace eckit::geo {
+
+
+class Cache {
+public:
+    using bytes_t = decltype(sizeof(int));
```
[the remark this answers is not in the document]
wdeconinck: C++11 section 5.3.3, paragraph 6 (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf):

> The result of sizeof and sizeof... is a constant of type std::size_t. [ Note: std::size_t is defined in the standard header <cstddef> (18.2). — end note ]

So `std::size_t` it is.

src/eckit/geo/grid/regular/RegularGaussian.cc
```diff
@@ -0,0 +1,85 @@
...
+}
+
+
+Grid::iterator RegularGaussian::cend() const {
+    return iterator{new geo::iterator::Regular(*this, size())};
+}
+
+
+Spec* RegularGaussian::spec(const std::string& name) {
+    auto N = Translator<std::string, size_t>{}(name.substr(1));
```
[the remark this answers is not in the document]
wdeconinck: Indeed: `std::stoull` (https://cplusplus.com/reference/string/stoull/)

src/eckit/geo/projection/PROJ.h
```diff
@@ -0,0 +1,123 @@
...
+    // -- Types
+
+    struct pj_t : std::unique_ptr<PJ, decltype(&proj_destroy)> {
+        using t = std::unique_ptr<PJ, decltype(&proj_destroy)>;
+        explicit pj_t(PJ* ptr) :
+            t(ptr, &proj_destroy) {}
+        explicit operator PJ*() const { return t::get(); }
+    };
+
+    struct ctx_t : std::unique_ptr<PJ_CONTEXT, decltype(&proj_context_destroy)> {
```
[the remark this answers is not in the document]
wdeconinck: Yes it can and should be done, as it is also done in Atlas: https://github.com/ecmwf/atlas/blob/develop/src/atlas/projection/detail/ProjProjection.h

src/eckit/maths/CMakeLists.txt
```diff
@@ -1,25 +1,26 @@
...
+    Lapack.h
     Matrix.h
     MatrixEigen.h
-    MatrixLapack.h)
+    MatrixLapack.h
+)
 
-if(eckit_HAVE_CONVEX_HULL)
-    list(APPEND eckit_maths_sources ConvexHull.h ConvexHullN.h Qhull.cc Qhull.h)
-    list(APPEND eckit_maths_private_libs Qhull::Qhull)
```
wdeconinck, line 12: Probably you did not mean to remove this.

tests/geo/iterator.cc
```diff
@@ -0,0 +1,28 @@
...
+
+
+#include "eckit/geo/Iterator.h"
+#include "eckit/testing/Test.h"
+
+
+namespace eckit::geo::test {
+
+
+CASE("") {}
```
wdeconinck, line 20: todo

tests/maths/CMakeLists.txt
```diff
@@ -1,16 +1,15 @@
...
+        SOURCES test_${_test}.cc
+        LIBS    eckit_maths eckit
+    )
+endforeach()
 
-ecbuild_add_test( TARGET      eckit_test_maths_convex_hull
-                  SOURCES     test_convex_hull.cc
-                  CONDITION   HAVE_CONVEX_HULL
-                  LIBS        eckit_maths eckit
-)
```
wdeconinck, line 16: You probably didn't mean to delete this.

src/eckit/geo/iterator/Reduced.cc
```diff
@@ -0,0 +1,95 @@
+/*
+ * (C) Copyright 1996- ECMWF.
```
wdeconinck, line 2: Copyright date here and in every file should be up to date I guess.

wdeconinck (changes requested):
I've managed to go through everything, although I cannot claim that I was super careful everywhere with such large PR.
I made all comments or suggestions in code.

My only real concern is the general Point class as std::variant
- the extra memory
- the (little) boiler plate needed to access the correct alternative
- the introspection needed to know (or overhead to assert) what's inside

I do understand the reason in creation of a more generic Projection API, or part of an iterator where the memory argument is mute, and I could definitely accept that there, but not when it is used in containers such as std::vector, or maps or trees, ...
I don't understand why UnstructuredFromGrid cannot be initialised with a vector of `PointLonLat` instead of `Point`, as that is what it is expecting. Also that means any grid could just have the function `to_pointlonlat` instead of `to_points` then, which reduces memory by half, possibly important at high resolution.

sandorkertesz:
@pmaciel , is there a gridspec json schema available to test the validity of the input gridspecs?

tests/geo/grid.cc
```diff
@@ -0,0 +1,103 @@
...
+            EXPECT_EQUAL(footprint_a, footprint_b);
+
+            const auto size_a = a->size();
+            const auto size_b = b->size();
+            EXPECT_EQUAL(size_a, size_b);
+        }
+    }
+
+
+    SECTION("Grid::build_from_increments") {
```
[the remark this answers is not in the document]
wdeconinck: Related, see issue https://github.com/ecmwf/eckit/issues/71 which mentions nested SECTIONS don't work. I also don't think it was intended to be possible in the first place.

src/eckit/geo/util/mutex.h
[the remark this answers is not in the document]
simondsmart: Yes, yes we should.

src/eckit/geo/Figure.h
pmaciel: Done

src/eckit/geo/Domain.h
pmaciel: Done

src/eckit/geo/Point.h
pmaciel: My commend there is that the specific resulting type depends on the configured Projection and Iterator for the containing Grid, so it is by design

src/eckit/geo/Point3.h
pmaciel: I've now refactored this taking into account the advice; There's an issue for const correctness because of brevity, but I've decided to enforce that points (all derived classes) are immutable to maintain internal consistency.

src/eckit/geo/Area.h
```diff
@@ -0,0 +1,118 @@
...
+    // -- Methods
+
+    static std::string className() { return "area"; }
+
+    virtual void spec(spec::Custom&) const;
+
+    // -- Overridden methods
+    // None
+
+    // -- Class members
```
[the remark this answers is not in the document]
pmaciel: I don't think they are a leftover from Baudouin, I have adopted his technique to prototype; It makes it easy to move methods and members up and down the hierarchy, as well as spotting design error, much easier. I even start with structs not classes to design, and then move onto this stage, and then clear when fixed "in stone"...

But I understand, will clean at lease the "None" labels...

src/eckit/geo/Cache.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "eckit/geo/Cache.h"
+
+#include <vector>
+
+
+namespace eckit::geo {
+
+
+static util::recursive_mutex MUTEX;
```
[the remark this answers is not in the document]
pmaciel: It's in the design; the several factories are based on self-registration, and the mutex is protecting the access on the generator methods (build, make) from the destruction, as well as locking out pre-main file reading so they happen only once. Also, for cache access to stop reading from if writing is taking place (the case here).

I use it blindly, they haven't posed a performance problem at scale; but, this should be revisited if there's resource contention at startup (itmight be the case already, but I'm not aware). One should also discuss applicability of other types of mutexes, but it's a super technical discussion that we aren't concerned about that I know of.

The fact I'm declaring util::recursive_mutex is to be honest to the developer -- internally, that's the kind of mutex (as to contrast from eckit::Mutex, which is in fact recursive).

src/eckit/geo/util/mutex.h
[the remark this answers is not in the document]
pmaciel: No this is the right name. I've made that exact change in mir only, and the whole stack fails in production (both in mars-client and pgen). This is even if we match the mutex type, but not the implementation (eckit::Mutex come from the system, while std::recursive_mutex interfaces directly pthreads); they aren't compatible, and the discussion far exceeds the scope here

src/eckit/geo/util/reduced_classical_pl.cc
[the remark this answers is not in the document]
pmaciel: Ah ah yes, this comes straight from Atlas; the main difference is the value_type is long (pl_type fordwards this type), which matches the GRIB format, hence no conversion takes place (so there should be a benefit here). If we put it in a source file it will pay the penalty in every instantiation, so no. Also, this is very small in footprint (I haven't calculated it but my estimate is... peanuts :-)

src/eckit/geo/iterator/Reduced.cc
pmaciel: This is up to date, as far as I understand we decided to label these

tests/maths/CMakeLists.txt
pmaciel: Indeed

src/eckit/geo/util/arange.cc
pmaciel: I'm trying that this is more future-proof, that we might gain the automatic parallelisation in future implementations (it's predicted, but I read it as science fiction). Compilers do a good job here too :-)

src/eckit/maths/CMakeLists.txt
pmaciel: No, it has more genral use, I was containing the changes limited to the library. I'll include this uncondittionally

src/eckit/geo/Cache.h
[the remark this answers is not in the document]
pmaciel: I could have easily used size_t, but that would be confusing for both the result of a generic cached type footprint and vector<T>, which are here specialised. This is an improvement of similar usage elsewhere (in eckit and below) where this isn't clear, consistent not concise. size_t is generic because it seems widespread usage for, among others, the result of containers size, an index and result of sizeof, neither of which mean the same thing at all.

Please read this way: the type is the result of calling sizeof, which should read clearly that I'm getting the result in bytes (hence the name, but bytes_size_t can be seen as an improvement). So "what is means" not "what it is".

Due to pouplar demand I'll call it bytes_size_t

src/eckit/geo/Projection.h
pmaciel: The semantics I use, and that was learned from experience on both the mars-client and pgen stack, is that if a function/method returns a pointer, it is up to the client code to catch it and manage it's life cycle. If the object is to be "used" but not "managed", then a reference suffices. It is very deliberate and throws a warning to any developer -- that they should care.

The use is typically from factories ("build") or type generators ("make"), and spec is the other case: it means that it generates an object of this type. I did consider std::unique_ptr would be useful with the spec generation, because it's the exception of the above, but decided to use the same pattern as a mix would be more confusing than clarifying.

But I agree yes, that with language modernization the [[nodiscard]] is mandatory with this pattern -- it won't be difficult to change everything in eckit::geo, but to change it in the stacks is a whole other problem...

tests/geo/grid.cc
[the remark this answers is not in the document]
pmaciel: Ok, I didn't knwo this was not working; I believe I might have written code like this in MIR or Atlas... I'm fixing it here for the moment

tests/geo/iterator.cc
pmaciel: Indeed :-)

pmaciel:
> @pmaciel , is there a gridspec json schema available to test the validity of the input gridspecs?

Not yet :-) But the unit tests should be a guide to write one. I've made the whole procedure automatic (eg., it derives a bounding box from a single coordinate, like just "west: 123" to north/west/south/east: 90/123/-90/123+360, or handles a octahedral reduced gg from a single key "N: 64")

We defintely should sit down and write a table of rules -- the ones I've written seem quite sufficient so far and are only maybe 5. This branch can already generate gridspec's if you build a Grid object, and the proof of fire will come first from mir (which will integrate nicelly with the "interpolationspec" already there), and afterwards from eccodes which will have a really comprehensive set of grids (historical, contemprary and wrongly encoded too! :-) )

pmaciel:
> I've managed to go through everything, although I cannot claim that I was super careful everywhere with such large PR. I made all comments or suggestions in code.
> 
> My only real concern is the general Point class as std::variant
> 
>     * the extra memory
> 
>     * the (little) boiler plate needed to access the correct alternative
> 
>     * the introspection needed to know (or overhead to assert) what's inside
> 
> 
> I do understand the reason in creation of a more generic Projection API, or part of an iterator where the memory argument is mute, and I could definitely accept that there, but not when it is used in containers such as std::vector, or maps or trees, ... I don't understand why UnstructuredFromGrid cannot be initialised with a vector of `PointLonLat` instead of `Point`, as that is what it is expecting. Also that means any grid could just have the function `to_pointlonlat` instead of `to_points` then, which reduces memory by half, possibly important at high resolution.

@wdeconinck Yes the idea here is to discuss the scope -- I believe the fundamental classes are written, and now I'm specialising (and fixing things of course). There's already been this week a "proof by fire" where we compared the performance of eccodes iterator to this one, and they are in the same ballpark (1%-2% performance difference either way), and the results matched down to several decimal places.

The Point class is fundamental here, yes. The point of using it as a variant and not inheritance is to: 1) benefit from the polymorphism without the virtual table penalty, 2) there's no interest in further specialising the point types (if you give it some thought), and 3) enforce checkling of types by the projections themselves, which actually need it specifically enforced (we would otherwise live with several overriden coordinates calculation functions, mostly empty or NOTIMP, and would still need to test dynamic_casts (which I'm alergic to). The cost here is a check on an integer, at the expense of inheritance which we don't need (but is still possible witha variant, but I see absolutely no interest in this library, maybe for projections?), and this check is both trivial and performant.

Performance also is not hindered by this choice: a spectral transfomrm will always win. In pgen the iterator is costly because the whole workflow is optimised, hence these numbers show there (I've measured it at 5%-8% a few years ago), but this library as really fast way to calculate grid sizes (which is what both MIR and eccodes actually needs).

There's a penalty but for performance concerns there's no issue, as these would never explicitly calculate vector<Point> (which yes incurrs on overhead) but rather use the iterator explicitly which is random accessso it can be present in OMP loops or MPI Ranks. The (larger) functions there are for prototyping tests, non-performance critical functions and interfacing with Python really.

pmaciel:
> This looks really nice overall, despite the number of small comments.
> 
>     * You seem to really like things declared with __ at the start. This is supposed to be reserved, and I think all of these can just be put into an anonymous namespace
> 
>     * Some of your types are returned with raw pointers. They either need [[nodiscard]], to return a std::unique_ptr, or to have a Value-like wrapper class that handles the pointer stuff internally.
> 
>     * There are a couple of bits (commented) where I don't think the changes are geo-specific. And they could do with going elsewhere. Particularly in spec, I wonder if we could do well to revisit the Parametrisation stuff in general.

@simondsmart thanks for reviewing! Yes my pattern is to start static variables with double underscore (just like in python, to stay clear); I've taken the advice and did the other Python way (or, the Fortran way, bless) all uppercase.

Yes @wdeconinck had the same concern; I don't find this a problem, and I agree with the semantics of returning naked pointers, like I said it is for the client code to care for it. But, the [[nodiscard]] is exactly for this like you mention, so I've changed. No Value-like classes please (yuck).

Yes the parametrisation is a challenge. I need to explicitly control the conversion of types, including vector types, and I cannot afford to rely on Value. -- it just isn't either performant, complete, consistent nor adequate for low level (eg. impossible to efficiently use over a grib handle). So I've written my own again based on a variant. It solves all these problems at the moment, and the only thing it cannot do is handle vectors of more than one type (we don't have this case here) nor nest configurations -- also not the case here, but trivial to add without again resorting the 2 levels of inheritance native to Value.

So there's the Spec class which supports these at the moment -- the challenge was and is ec codes. This implementation is currently adequate, but feel free to dig deeper, I'm sure the work here is not finished, nor is the a solution for everything, just for eckit::geo.

src/eckit/geo/Cache.h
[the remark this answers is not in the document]
wdeconinck: For me it was not so much the naming `bytes_t` vs. `bytes_size_t` vs `size_t` but rather the unnecessary `decltype`.
I would have liked `using bytes_t = std::size_t` as the standard is explicit about the return type of `sizeof()`.
It is just a minor thing which I don't care about so "whatever" :)
Also note that C++17 introduced `std::byte` to represent a single byte. Completely different use but naming is similar.

iainrussell:
@pmaciel - _not to be looked at until next week_, but I enabled this feature (ECKIT_GEO) on the downstream-ci, which is why the builds are now failing - compilation errors on different platforms.

I temporarily enabled PROJ, but have now disabled it because none of our Linux runners have version 9.2 installed (nor is it available for any of the distributions). I checked, and here are the versions of proj that we have installed on the runners (note: many of these are quite old distros now, and near end of maintenance, so we may drop them soon). Probably most of the proj versions could be accommodated, but the one in centos 7.9 could be a problem...

proj    os
8.2   ubuntu 22
7.2   debian 11
4.2   centos 7.9
6.3   rocky 8.6
9.0   Fedora 37

wdeconinck:
What is the meaning of PointLonLatR ?  What is the difference with PointLonLat? Why is it needed?
Also please document in code.

PR #117, opened 2024-04-17
fix PooledFile multi threading

danovaro:


src/eckit/io/PooledFile.cc
```diff
@@ -26,7 +26,8 @@ namespace eckit {
 
 class PoolFileEntry;
 
-static thread_local std::map<PathName, std::unique_ptr<PoolFileEntry>> pool_;
+static std::map<PathName, std::unique_ptr<PoolFileEntry>> pool_;
```
tlmquintino, line 29: please put this in a class with a static method with static variable as we have in there parts of the codebase.
this will ensure correct order of initialisation on use, not related to linking order. 
the mutex is then a member of that class that you can lock and unlock as you insert and remove entries to the map via proper defined methods

src/eckit/io/PooledFile.cc
```diff
@@ -73,10 +74,11 @@ class PoolFileEntry {
 
     void remove(const PooledFile* file) {
         auto s = statuses_.find(file);
-        if (s != statuses_.end()) {
-            statuses_.erase(s);
-        }
+        ASSERT(s != statuses_.end());
```
tlmquintino, line 77: this changes the behaviour slightly, but looks better and ensuring stronger correctness.

tlmquintino (changes requested):
please address my first comment

src/eckit/io/PooledFile.cc
```diff
@@ -21,12 +21,31 @@
 #include "eckit/io/PooledFile.h"
 #include "eckit/log/Bytes.h"
```
pmaciel, line 23: missing includes map, mutex, utility, sstream
remove thread
move #include "eckit/io/PooledFile.h" to the top of the include list

src/eckit/io/PooledFile.cc
```diff
@@ -262,3 +275,21 @@ PooledFileError::PooledFileError(const std::string& file, const std::string& msg
...
 
 }  // namespace eckit
+
+namespace {
+
+eckit::PoolFileEntry* Pool::get(const eckit::PathName& name) {
+    std::lock_guard<std::mutex> lock(filePoolMutex_);
+    auto j = filePool_.find(name);
+    if (j == filePool_.end()) {
+        filePool_.emplace(std::make_pair(name, std::unique_ptr<eckit::PoolFileEntry>(new eckit::PoolFileEntry(name))));
```
pmaciel, line 285: remove the std::make_pair, you can emplace(key, value) directly avoiding a (actualy, more than one) temporary

src/eckit/io/PooledFile.cc
```diff
@@ -113,10 +132,10 @@ class PoolFileEntry {
...
     void close(const PooledFile* file) {
         auto s = statuses_.find(file);
-        if (s != statuses_.end()) {
-            ASSERT(s->second.opened_);
-            s->second.opened_ = false;
-        }
+        ASSERT(s != statuses_.end());
+
+        ASSERT(s->second.opened_);
+        s->second.opened_ = false;
```
pmaciel, line 138: Excellent

src/eckit/io/PooledFile.cc
```diff
@@ -21,12 +21,31 @@
...
+}
+
+namespace {
+
+class Pool {
+
+private:
+    Pool() {}
+    std::map<eckit::PathName, std::unique_ptr<eckit::PoolFileEntry>> filePool_;
+    std::mutex filePoolMutex_;
```
pmaciel, line 35: You just have to be careful not missing threading models, std::mutex and eckit::Mutex (a recursive mutex based on pthread_mutex_t)

pmaciel (approved):
I'm only commenting on the changed code, I don't think anything changed significantly other than the explicit use of std::mutex -- which I assume is common to downstream code, as I know there's explicit use of eckit::Mutex in some dowstream packages.

simondsmart (approved):

tlmquintino (commented):
All good now.

PR #118, opened 2024-04-23
Use std::string_view in eckit_mpi

wdeconinck:
This PR changes `const char*` arguments into `std::string_view`, so that we can now downstream replace amongst others
```eckit::mpi::setCommDefault( string.c_str() )``` and ```eckit::mpi::comm( string.c_str() )```
with 
```eckit::mpi::setCommDefault( string )``` and ```eckit::mpi::comm( string )```.

It should still be compatible with callers passing `const char*`.

### DIFF
--- a/src/eckit/mpi/Comm.cc
+++ b/src/eckit/mpi/Comm.cc
@@ -36,7 +36,7 @@ constexpr bool have_parallel() {
 
 class Environment {
 public:
-    static const char* getDefaultCommType() {
+    static std::string_view getDefaultCommType() {
         // Force a given communicator (only required if e.g. running serial applications with MPI)
         if (const char* forcedComm = ::getenv("ECKIT_MPI_FORCE")) {
             return forcedComm;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/eckit/mpi/Comm.cc [hunk 1]
```diff
@@ -95,10 +95,10 @@ class Environment {
         for (itr = communicators.begin(); itr != communicators.end(); ++itr) {
             eckit::Log::error() << "   " << (*itr).first << std::endl;
         }
-        throw eckit::SeriousBug(std::string("No communicator called ") + name, Here());
+        throw eckit::SeriousBug(std::string("No communicator called ") + std::string(name), Here());
```
pmaciel, line 98: Being pedantic here, for even more terseness one can remove the first std::string construction (in more than one place I see)

pmaciel (approved):
Aside from my minor comment, I only add that there is the function MPICall (Parallel.cc, and two fwd declarations in ParallelGroup.cc and ParallelRequest.cc) which takes const char* which might by more consistently converted also to std::string_view. I'm only advancing this if it was missed, I have no specific opinion or preference here.

src/eckit/mpi/Comm.cc
dsarmany: Yeah, that would still work. My nitpick comment is that I prefer braces '{}' for calling the constructor to make it immediately obvious that we are calling the constructor and not a free function. This used to be very helpful for not getting caught out by 'the most vexing parse' behaviour, but these days compilers would warn you about that anyway.

Otherwise, it looks good.

dsarmany (approved):

src/eckit/mpi/Comm.cc
wdeconinck: Thanks! We aim for perfection :) I have implemented all suggestions.

PR #120, opened 2024-05-08
feat(CI): added MS Teams notifications

mcakircali:


geier1993 (approved):
Looks fine to me. Have modified the conditions because there seems to be a problem with `always()` is excluded from ${{ ... }}. 
With `always()` the action is also triggered even if the dependencies (`needs`) fail.
By writing `if: always() && ${{ ... }}` this was not recognized properly

PR #122, opened 2024-05-28
Feature/syslog structured data

sametd:
This PR implements the "origin" component of the syslog structural data as described in RFC5424, specifically in sections 6.2 and 7.2 ([RFC5424 Section 7.2](https://datatracker.ietf.org/doc/html/rfc5424#section-7.2)). Implementation also ensures that the current functionality remains unaffected if the newly introduced variables are not set. I am tagging @tlmquintino here as he may also want to take a look at this implementation.

Note: This is not a full implementation of structural data of the syslog, only implements the bits that necessary to add logserver-compatible logging. Logserver repo can be found [here](https://github.com/ecmwf/logserver-ansible)

### DIFF
--- a/src/eckit/log/SysLog.h
+++ b/src/eckit/log/SysLog.h
@@ -99,6 +99,11 @@ public:  // methods
         return s;
     }
 
+    /// Optional fields for structured data (RFC 5424 section 6.3)
+    void software(const std::string& software) { software_ = software; }
+    void swVersion(const std::string& version) { swVersion_ = version; }
+    void enterpriseId(const std::string& id) { enterpriseId_ = id; }
+
 private:  // methods
     void print(std::ostream& out) const;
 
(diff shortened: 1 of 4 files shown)


src/eckit/log/SysLog.h
```diff
@@ -99,6 +99,11 @@ class SysLog {
         return s;
     }
 
+    /// Optional fields for structured data (RFC 5424 section 6.3)
+    void software(const std::string& software) { software_ = software; }
```
simondsmart, line 103: Where are these being set?

simondsmart:
Looks uncontroversial. Why is there a failing test?

src/eckit/log/SysLog.h
sametd: These are the keywords that belongs to origin part of the syslog structural data described in RFC5424 -> https://datatracker.ietf.org/doc/html/rfc5424#section-7.2

simondsmart (approved):
Looks good to me. Once the CI is happy, we should be good to go.

PR #123, opened 2024-05-28
Adds new backend OpenFAM

mcakircali:


src/eckit/io/fam/FamConfig.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace eckit {
+
+//----------------------------------------------------------------------------------------------------------------------
+
+struct FamConfig {
+    net::Endpoint endpoint {"127.0.0.1", -1};
+    std::string   sessionName {"EckitFamSession"};
+
+    bool operator==(const FamConfig& other) const {
```
geier1993, line 35: Shouldn't that be put in a CC file?
Otherwise if included in different places the compiler might complain that the operator is already defined.

src/eckit/io/fam/FamListIterator.h
```diff
@@ -0,0 +1,78 @@
...
+    using iterator_category = std::bidirectional_iterator_tag;
+    using value_type        = FamObject;
+    using pointer           = FamObject*;
+    using reference         = Buffer&;
+
+public:  // methods
+    FamListIterator(const value_type& object);
+
+    // iterate forwards
+    auto operator++() -> FamListIterator&;
```
geier1993, line 41: No need for trailing return types. ok for me, just as a note

src/eckit/io/fam/FamListIterator.h
```diff
@@ -0,0 +1,78 @@
...
+
+    // iterate backwards
+    auto operator--() -> FamListIterator&;
+
+    auto operator==(const FamListIterator& other) const -> bool;
+
+    auto operator!=(const FamListIterator& other) const -> bool { return !operator==(other); }
```
geier1993, line 48: definition should be put in CC file ?

src/eckit/io/fam/FamListIterator.h
```diff
@@ -0,0 +1,78 @@
...
+class FamListConstIterator: public FamListIterator {
+    using FamListIterator::FamListIterator;
+
+    using pointer   = const FamObject*;
+    using reference = const Buffer&;
+
+public:  // methods
+    auto operator->() -> pointer { return FamListIterator::operator->(); }
+
+    auto operator*() -> reference { return FamListIterator::operator*(); }
```
geier1993, line 73: put definitions in CC file?

src/eckit/io/fam/FamPath.h
```diff
@@ -0,0 +1,60 @@
...
+
+    FamPath() = default;
+
+    FamPath(const std::string& path);
+
+    FamPath(const char* path): FamPath(std::string(path)) { }
+
+    FamPath(const URI& uri);
+
+    auto generateUUID() const -> std::string;
```
geier1993, line 42: no need for trailing return type

src/eckit/io/fam/FamPath.h
```diff
@@ -0,0 +1,60 @@
...
+
+    bool operator==(const FamPath& other) const {
+        return (regionName == other.regionName && objectName == other.objectName);
+    }
+
+    auto asString() const -> std::string {
```
geier1993, line 48: no need for trailing return type

can also create an `operator std::string` for implicit or explicit only conversion

src/eckit/io/fam/FamObject.h
```diff
@@ -0,0 +1,124 @@
...
+    auto compareSwap(fam::size_t offset, T oldValue, T newValue) const -> T;
+
+private:  // methods
+    void print(std::ostream& out) const;
+
+    friend std::ostream& operator<<(std::ostream& out, const FamObject& object);
+
+private:  // members
+    std::shared_ptr<FamSessionDetail>    session_;
+    std::shared_ptr<FamObjectDescriptor> object_;
```
geier1993, line 119: Is the heavily use of `shared_ptr` intentional??
Are there no clean ownerships or live time assumptions when returning these?

src/eckit/io/fam/FamRegion.cc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+auto FamRegion::property() const -> FamProperty {
+    return {size(), permissions(), name()};
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+// OBJECT factory methods
+
+auto FamRegion::proxyObject(const std::uint64_t offset) const -> FamObject {
```
geier1993, line 84: Can't the FamObject be returned as reference or const reference??

I don't know how it is designed

geier1993 (commented):
Have just reviewed code, can not judge about overall logic.

Think it's quite clean code.
Stylistically heavily using trailing return types is different not needed but also no disadvantage. Actually, IMO it is more readable.

The use of the shared_ptr may need some more discussion here. Just knowing that @simondsmart prefers not to use shared_ptr when ownership is clean (or can be made clean)
and object lifetime is no matter.

src/eckit/io/fam/FamObject.h
mcakircali: yes. shared_ptr because;
1- sessions has to be "shared" since they connect to a remote; and we re-use that connection.. (some backends such as aws "require" single client sessions)
2- FamObject needs copy semantics
3- shared_ptr + unique_ptr would be a non-meaningful semantics

src/eckit/io/fam/FamConfig.h
mcakircali: no. this is operator== overloading inside FamConfig .. 

"inline bool eckit::FamConfig::operator==(...)"

notice implicit inline as well as its name space

src/eckit/io/fam/FamListIterator.h
mcakircali: could be, but this particular definition will never ever change (to break any API later)

src/eckit/io/fam/FamRegion.cc
mcakircali: no, it would not make sense. we don't need nor want object book keeping

mcakircali:
closing to separate Buffer issue from this

PR #134, opened 2024-06-21
Conditional CI notify step

mcakircali:
Limit CI MS Teams notifications to: failure at master/main/develop


wdeconinck (approved):
Ok to me

PR #135, opened 2024-06-21
Add isNull to eckit::Configuration

geier1993:
I would like to check properly if a value ina yaml configuration is NULL.
Otherwise I can just guess that a value is null by exclusion (when all the other introspective calls are false).

### DIFF
--- a/src/eckit/config/Configuration.cc
+++ b/src/eckit/config/Configuration.cc
@@ -500,6 +500,12 @@ bool Configuration::isSubConfiguration(const std::string& name) const {
     return found && (v.isMap() || v.isOrderedMap());
 }
 
+bool Configuration::isNull(const std::string& name) const {
+    bool found     = false;
+    eckit::Value v = lookUp(name, found);
+    return found && (v.isNil());
+}
+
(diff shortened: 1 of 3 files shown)


pmaciel (approved):
This makes sense; Note, the current eckit::Value instrospection methods list:
isBoolean
isBooleanList
isFloatingPoint
isFloatingPointList
isIntegral
isIntegralList
isList
isString
isStringList
isSubConfiguration
isSubConfigurationList
isNull (new)

is not very similar to the eckit::Content:
isBool
isDate
isDateTime
isDouble
isList
isMap
isNumber
isOrderedMap
isString
isTime
isNil

neither to the standard <type_traits>, which I believe that's where we should head. But until this is changed, this is a improvement :smiley::+1:

wdeconinck (approved):

wdeconinck:
Just one more thing… a unit test 🤓

geier1993:
> Just one more thing… a unit test 🤓

added a test

dsarmany:
Only the multio-python CI fails which has nothing to do with this change and is being looked into.

PR #136, opened 2024-07-08
Adds fabric attached memory support

mcakircali:
FAM support via OpenFAM


src/eckit/io/Buffer.h
```diff
@@ -40,6 +40,8 @@ class Buffer : private NonCopyable {
 
     ~Buffer();
 
+    std::string_view view() const noexcept { return std::string_view(buffer_, size_); }
+
```
Ozaq, line 44: `Buffer` is a byte buffer and while it could be used as a c-string it think this should not be done. Especially. since the only usage I can see is in tests.

src/eckit/io/fam/FamConfig.h
```diff
@@ -0,0 +1,47 @@
...
+    std::string   sessionName {"EckitFamSession"};
+
+    bool operator==(const FamConfig& other) const {
+        return (endpoint == other.endpoint && sessionName == other.sessionName);
+    }
+
+    friend std::ostream& operator<<(std::ostream& out, const FamConfig& config) {
+        out << "endpoint=" << config.endpoint << ", sessionName=" << config.sessionName;
+        return out;
+    }
```
Ozaq, line 42: This does not need to be a friend.

PR #137, opened 2024-07-08
Buffer zero size alloc

mcakircali:
In eckit::Buffer default ctor (size=0), buffer_ is allocated as 'new char[0]'.

'buffer.data() != nullptr', which is not a correct state for an uninitialized buffer


wdeconinck (approved):
Looks uncontroversial to me :)

PR #140, opened 2024-09-20
Add support for colon seperated list of library homes

ChrisspyB:
Allows setting multiple homes in a colon seperated string. e.g. `ECKIT_HOME=/path/to/home1:/path/to/home2`
This is useful for libraries which like to search for multiple configs, which may be in distinct directories. LocalPathName's tilde expansion `~eckit/file` will match the first existing file from `[/path/to/home1/file1, /path/to/home2/file]`. If none exist, the first is used.

No change in behaviour is intended if a single home is set.

If LIB_HOME is not set, there is a change in behaviour in the tilde expansion:
 - Before: eckit would search recursively upwards from the prefix directory all the way to the root directory.
   - e.g. `~eckit/file1` -> `libraryprefix/file1`, `libraryprefix/../file1`, `libraryprefix/../../file1`, `libraryprefix/../../../file1`... `/file1`
 - Now: eckit will search the prefix directories and one and two levels above it, then go straight to the root. (exactly four directories)
    - e.g. `~eckit/file1` -> `libraryprefix/file1`, `libraryprefix/../file1`, `libraryprefix/../../file1`, `libraryprefix/../../file1`, `/file1`

**note** This does mean that LIB_HOME cannot be set to directories whose path contains a colon.

### DIFF
--- a/src/eckit/system/Library.cc
+++ b/src/eckit/system/Library.cc
@@ -74,29 +74,42 @@ std::string Library::prefixDirectory() const {
     return prefixDirectory_;
 }
 
-std::string Library::home() const {
+const std::vector<std::string>& Library::homes() const {
     AutoLock<Mutex> lock(mutex_);
 
-    std::string libhome = prefix_ + "_HOME";
-    char* home          = ::getenv(libhome.c_str());
-    if (home) {
-        return home;
+    if (!homes_.empty()) {
+        return homes_;
     }
 
-    return home_;  // may return empty string (meaning not set)
+    std::string libhome = prefix_ + "_HOME";
+    char* home_env = ::getenv(libhome.c_str());
+
+    // Home is a colon separated list of directories
+    if (home_env) {
+        std::stringstream ss(home_env);
+        std::string item;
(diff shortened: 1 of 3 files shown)


ChrisspyB:
Hello past me.
I think this would be nice to have. I should re-review what I wrote a year ago and then push someone else to review/merge.

ChrisspyB:
Hi @pmaciel. I will have to review this myself too as it was a long time ago, but from what I remember.

> what happens if configuration files of the same name are found in multiple paths

The intention is the first found path takes precedence. 

> are non-existant paths allowed

Yes, as these are allowed with the current implementation. In these cases we resolve to the first home. e..g the `~home/not_existent_file` will resolve to `/path/to/first_home/non_existent_file`, if the file is not present in any of the home directories.

> ie., separately PROJECT_A_HOME=/this and PROJECT_B_HOME=/that, then PROJECTS_A_B_HOME=/this:/that

I was completely unaware of this functionality, and it seems very likely that these changes would break it... (e.g. what if I set 2 homes for A and 2 homes for B...?).

PR #141, opened 2024-09-27
Fix compilation warnings on LUMI with cce/17

wdeconinck:
This fixes a few warnings when compiling eckit on LUMI with CrayClang compiler cce/17.
One warning was for unrecognised flag for CrayClang as opposed to "just" Cray.
Another related to deprecated implicit assignment operator.

### DIFF
--- a/src/eckit/codec/Stream.h
+++ b/src/eckit/codec/Stream.h
@@ -46,8 +46,11 @@ public:
     ///       the referenced datahandle
     Stream(DataHandle&);
 
-    /// Assignment constructor sharing datahandle with other Stream
-    Stream(const Stream&);
+    /// Assignment/Copy constructor sharing datahandle with other Stream
+    Stream(const Stream&) = default;
+    Stream(Stream&&) = default;
+    Stream& operator=(const Stream&) = default;
+    Stream& operator=(Stream&&) = default;
 
     /// Access internal DataHandle
     DataHandle& datahandle();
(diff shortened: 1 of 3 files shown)


src/eckit/codec/Stream.h
mcakircali, line 1: this change adds `move` semantics, ctor and assignment.
previously, it was `only copy`
which is unusual for a class with a shared_ptr member

mcakircali: using `move` semantics (3 and 4) on a class with a member of shared_ptr type looks unusual to me

usual use of shared_ptr type objects is to copy it around, which increments count; not move it around

mcakircali: so, it is a big change of how this class works. 
I like (and also prefer) the explicit definitions. it tells others it's intentional and clear.
I suggest, move ctor and assignment `= delete;`

wdeconinck: Sure, I will remove it. I just added it for completeness as the rule-of-five.

wdeconinck: Marking it with `=delete` does not work and causes further errors. So I suggest leaving it with `=default` marks that this is the intended and supported behaviour.

wdeconinck:
This PR has now gone through all the tests successfully.

mcakircali (approved):
approved as it silences warnings
moved-from obj leaks pointer and shall be fixed in another PR

PR #142, opened 2024-10-11
perf fix Time

mcakircali:
try to avoid `std::regex`


src/eckit/types/Time.cc
```diff
@@ -67,60 +98,56 @@ Time::Time(const std::string& s, bool extended) {
...
+            hh = (ss / 3600) % 24;
+            mm = (ss / 60) % 60;
+            ss = ss % 60;
+        } else {
+            throw BadTime("Wrong input for time: " + s);
         }
+    } else if (format == TimeFormat::UNKOWN) {
+        throw BadTime("Unkown format for time: " + s);
+    } else {
+        throw SeriousBug("Unhandled time format!");
```
Ozaq, line 150: Why is there a need for two different exceptions? I understand the intent of those exception types as follows: 
- `BadTime` is best described as a usage error, as in it was called as `Time("ABCD")`
- `SeriousBug` represents an implementation error such as `Time("20:11:24")` not getting parsed properly.

Is that second throw giving any additional benefits?

src/eckit/types/Time.cc
```diff
@@ -8,34 +8,64 @@
...
 
 //----------------------------------------------------------------------------------------------------------------------
 
-inline void printTime(std::ostream& s, long n) {
+const std::regex hhmmss_("^([0-9]+):([0-5]?[0-9])(:[0-5]?[0-9])?$");
+const std::regex ddhhmmss_("^-?([0-9]+[dD])?([0-9]+[hH])?([0-9]+[mM])?([0-9]+[sS])?$");
+
+// DIGITS: "^-?[0-9]+$"
+// FLOAT:  "^-?[0-9]*\\.[0-9]+$"
+enum class TimeFormat { UNKOWN, OTHER, DIGITS, DECIMAL };
```
Ozaq, line 32: `UNKNOWN` is treated as erroneous input, it go stronger in the name and call it `INVALID`

src/eckit/types/Time.cc
```diff
@@ -67,60 +98,56 @@ Time::Time(const std::string& s, bool extended) {
...
+    } else if (format == TimeFormat::DECIMAL) {
+        long sec  = std::round(std::stod(s) * 3600);
+        hh   = sec / 3600;
+        sec -= hh * 3600;
+        mm   = sec / 60;
+        sec -= mm * 60;
+        ss   = sec;
+    } else if (format == TimeFormat::OTHER) {
+        std::smatch m;
+        if (std::regex_match(s, m, hhmmss_)) {
```
Ozaq, line 110: This is still using a regex.

src/eckit/types/Time.cc
```diff
@@ -67,60 +98,56 @@ Time::Time(const std::string& s, bool extended) {
...
-                                case 2: hh = t; break;
-                                case 3: mm = t; break;
-                                case 4: ss = t;
-                            }
-                        }
-                    }
-                    ss += 60 * (mm + 60 * (hh + 24 * dd));
-                    if (s[0] == '-') {
-                        ss = -ss;
+        } else if (std::regex_match(s, m, ddhhmmss_)) {
```
Ozaq, line 124: This is still using a regex.

src/eckit/types/Time.cc
```diff
@@ -8,34 +8,64 @@
...
+    for (auto i = start; i < time.length(); i++) {
+        if (time[i] == '.') {
+            if (hasDecimal || i == time.length() - 1) { return TimeFormat::UNKOWN; }
+            hasDecimal = true;
+        } else if (isdigit(time[i]) == 0) {
+            return TimeFormat::OTHER;
+        } else {
+            hasDigit = true;
+        }
+    }
```
Ozaq, line 49: I think this can be done with a single pass over the input string and then return the typeid of the found variant (pretty similar to what you are doing already) plus up to 5 tokens, 1 token for the sign and 4 for positive integers (day, hour, minute, second).

Subsequent code can then validate on this result, e.g. was the extended flag passed.

I am thinking along the lines of:
```cpp
struct tokenized_time {
    FormatType type;
    std::string_view sign;
    std::string_view integers[4];
};
```

Ozaq (commented):
While you are at this class, I would also encourage you to add some class level doc and describe the accepted format variants and to be expected exceptions

PR #143, opened 2024-10-18
Fix missing header in ResourceUsage.h

SF-N:
`ResourceUsage.h` used `size_t`, but did not include the stddef header. This lead to some compilation errors in atlas4py.

### DIFF
--- a/src/eckit/system/ResourceUsage.h
+++ b/src/eckit/system/ResourceUsage.h
@@ -18,6 +18,7 @@
 #include <sys/resource.h>
 #include <cstddef>
 #include <iosfwd>
+#include <stddef.h>
 
 namespace eckit::system {
 


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=143) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=143) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/eckit?pullRequest=143) it.</sub>

pmaciel:
Note that the #include <cstddef> above takes care of this

SF-N:
It wasn't working with clang-1500.3.9.4 and when looking at https://en.cppreference.com/w/c/types/size_t it seems like the includes above don't take care of this.

wdeconinck:
Something strange, because you should not use the C compiler but rather the C++ compiler, and there `size_t` should be defined in many possible headers including `cstddef` as is already included.
"https://en.cppreference.com/w/cpp/types/size_t"

Please could you provide us the platform information, full compiler invocation and error message?

SF-N:
@wdeconinck and I just figured it out: it was due to an older eckit-version used in atlas4py which didn't include that header.

PR #145, opened 2024-10-24
Update FindAEC and add to CI

ChrisspyB:
Three changes:
- Report `dlerror` on a failed `dlopen` call.
- Update findAEC.cmake to be consistent with `eccodes`' findAEC.cmake. This searches more paths, and also looks for libaec.h rather than libsz.h
- Add AEC to the CI dependencies so that it uses a recent AEC, rather than whatever happens to be on the runners' systems.

Context:
Downstream dependency `gribjump` generates a plugin that depends on libaec/1.1.1 which is not backwards compatible. However, if eckit is linked against the system AEC (which is generally quite old), this old AEC will be loaded before the AEC gribjump is linked against. As such, whenever eckit loads the gribjump plugin at runtime, there is a missing symbol error. 

This PR updates the CI to build AEC from source, and updates the findAEC search path to find it. We also now throw an error in `loadPlugin` if `loadDynamicLibrary` returns nullptr. Without this, an incorrect error message is thrown instead (plugin object not registered).

### DIFF
--- a/.github/ci-config.yml
+++ b/.github/ci-config.yml
@@ -1,5 +1,6 @@
 cmake_options: -DENABLE_ECKIT_GEO=ON
 dependencies: |
   ecmwf/ecbuild
+  MathisRosenhauer/libaec@master
 dependency_branch: develop
 parallelism_factor: 8
--- a/cmake/FindAEC.cmake
+++ b/cmake/FindAEC.cmake
@@ -18,14 +18,19 @@
 #
 #  AEC_DIR          - prefix path of the AEC installation
 #  AEC_PATH         - prefix path of the AEC installation
+#  LIBAEC_DIR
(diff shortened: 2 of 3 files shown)


ChrisspyB:
It might also be sensible to put findAEC.cmake in ecbuild, as there are now several packages with their own implementation

wdeconinck:
> It might also be sensible to put findAEC.cmake in ecbuild, as there are now several packages with their own implementation

I agree that if this is goes into multiple packages then ecbuild could be a good common place.

wdeconinck (approved):
It all looks good to me!

iainrussell (approved):
Looks good to me too!

ChrisspyB:
If you are happy, please go ahead and merge :)

PR #146, opened 2024-10-28
buffer zero alloc

mcakircali:


wdeconinck:
Please could you provide a description to the issue and what this is solving?

mcakircali:
> Please could you provide a description to the issue and what this is solving?

done

wdeconinck (approved):
This looks like a good and sensible change to me.

Ozaq (commented):
Looking at Buffer this could be simplified by using `std::vector<>char` as backing container. 

This would remove some code such as allocate/deallocate 

Further It would be good to add documentation to the public interface since you already touched this class. Construction from std::string and copy from std::string deserve a short hint in the api doc imo as they treat std::string as a null terminated c-string. i.e. the resulting buffer length is string.size() + 1.

FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=146) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/eckit?pullRequest=146) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: mcakircali<br/>:x: Ozaq<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/eckit?pullRequest=146) it.</sub>

Ozaq:
pushed doc after talking to @mcakircali

pmaciel:
Hi @mcakircali , I went ahead and made small modernisation changes, logic stays the same. If you disagree feel free to reverse, otherwise I believe this is a good change and ready to go (for some time now).

@Ozaq agree with your observation of deallocations/allocations with a vector instead as a buffer container. The only downside I see is that vector has copy semantics (this Buffer doesn't have these), and typically buffers aren't copied -- runtime performace comes from staticly allocated, reusable buffers. To me I'm 49% fwd, 51% against, it's not a perfect fit.

pmaciel (approved):

wdeconinck:
I disagree to use std::vector as a backing as it calls the default constructor for each vector element. Usually that sets POD to zero. The performance loss is significant.

Ozaq:
> I disagree to use std::vector as a backing as it calls the default constructor for each vector element. Usually that sets POD to zero. The performance loss is significant. So this is good as-is.

I don't want to argue for using a std::vector but I think this overstates the impact

We have a `std::vector<char>` so there are no individual constructors called, but the memory is value initialized which should result in a alloc + memset, true there is additional cost to memset the newly allocated memory to zero.

geier1993:
> pushed doc after talking to @mcakircali
> 
> Edit: Also the Buffer dtor should be virtual as this class is subclassed and has protected members. I think making this class final would be better but that chan get the API


That's just a small but important change, we may add it before we merge?

mcakircali:
> > pushed doc after talking to @mcakircali
> > Edit: Also the Buffer dtor should be virtual as this class is subclassed and has protected members. I think making this class final would be better but that chan get the API
> 
> That's just a small but important change, we may add it before we merge?

this PR doesn't intend to fix inheritance issues nor missing documentation.

such fixes should go in their own (separate) PR that can be reviewed and merged more quickly than waiting this PR.

PR #148, opened 2024-11-13
Feature/eckit sparsematrix allocators reorganization, new in-place allocator

pmaciel:
This add a new "allocator" to the SparseMatrix, which allows building one with direct, non-owned access to a set of Index*, Index* and Scalar*.

It allows low-level control of the data structures, benefiting the following:
- to provide views, hence promoting parallelism
- efficient handling of very large cases, by exploiting the internal structure (per-row assembly, non-linear corrections, etc.)
- creation of utilities that observe and manipulate the structure directly (reordering, manipulation of weights, etc.)

There's no urgency

### DIFF
--- a/src/eckit/linalg/allocator/InPlaceAllocator.cc
+++ b/src/eckit/linalg/allocator/InPlaceAllocator.cc
@@ -1,0 +1,57 @@
+/*
+ * (C) Copyright 2024- ECMWF.
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation
+ * nor does it submit to any jurisdiction.
+ */
+
+
+#include "eckit/linalg/allocator/InPlaceAllocator.h"
+
+#include <ostream>
+
+#include "eckit/exception/Exceptions.h"
+
+
+namespace eckit::linalg::allocator {
+
+
+InPlaceAllocator::InPlaceAllocator(Size Nr, Size Nc, Size nnz, Index* ia, Index* ja, Scalar* a) :
+    Nr_(Nr), Nc_(Nc), nnz_(nnz), ia_(ia), ja_(ja), a_(a) {
+    ASSERT(ia_ != nullptr);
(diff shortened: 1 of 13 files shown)


pmaciel:
Say a few more days before merge? I have changes depending on this.

src/eckit/linalg/allocator/InPlaceAllocator.cc
```diff
@@ -0,0 +1,57 @@
...
+}
+
+
+SparseMatrix::Layout InPlaceAllocator::InPlaceAllocator::allocate(SparseMatrix::Shape& shape) {
+    shape.size_ = nnz_;
+    shape.rows_ = Nr_;
+    shape.cols_ = Nc_;
+
+    SparseMatrix::Layout layout;
+    layout.outer_ = reinterpret_cast<decltype(SparseMatrix::Layout::outer_)>(ia_);
```
wdeconinck, line 36: I believe a reinterpret_cast or const_cast or any other casting should not be necessary.

wdeconinck (changes requested):
I think this is great! A few points.

minor point: I would change to something like "NonOwningAllocator" or "WrappingAllocator"

more significant point:
When the SparseMatrix is created this way, wrapping existing data, the user of this class should not be allowed to call several methods:
- `SparseMatrix::reserve(rows, cols, nnz)`
- `SparseMatrix::load()`
- `SparseMatrix::transpose()`
- `SparseMatrix::setIdentity()`
- `SparseMatrix::prune()`

In fact anything that is attempting to change the shape should be forbidden.

One option is to add a boolean flag to `SparseMatrix::Shape` be set by the Allocator which can be checked if changes to shape are allowed or not. Open to other suggestions.

src/eckit/linalg/allocator/InPlaceAllocator.cc
wdeconinck: I understood now that the reinterpret_cast is a byproduct of a PR from a few months back where the data type of outer_ has changed from Index to UIndex.

wdeconinck (approved):
OK I agree that it is ambiguous what needs to happen with e.g. the `matrix.transpose()` case when using the non-owning allocator. It had bitten me that the wrapped memory became in a corrupt or undefined state because of such function, and without any warning.

This can be left for future pain and frustration ¯\_(ツ)_/¯ 
I'm OK for this to be merged now.

PR #154, opened 2024-12-09
Feature/eckit stats

pmaciel:
Mirco, I've ported 90% of the mir statistics functionality here, it's not a full PR to be merged before much discussion but rather a draft.

For MIR, the requirements were:
- supporting finite and NaN missing values;
- Calculate statistics in time (for weather statistics and climatotlogical studies) and/or in space (supporting a variety of interpolation methods);
- Do online statistics, in partitioned objects that can be summed for global results (for parallel execution, although this was never taken advantage off);
- Support comparisons for pgen (which is a lot); And
- To support eccodes namespace "statistics" (grib_ls -n statistics), which is why there are spectral statistics too, but this isn't a requirement

I'm sure for other packages it will be different, let's see if they converge to something useful :-)


pmaciel:
Again please do not merge this, before we have a wider discussion with other eckit devs.

src/eckit/stats/Statistics.h
```diff
@@ -0,0 +1,97 @@
...
+    Statistics(const Parametrisation&);
+    Statistics(const Statistics&) = delete;
+
+    // -- Destructor
+
+    virtual ~Statistics();
+
+    // -- Operators
+
+    void operator=(const Statistics&) = delete;
```
simondsmart, line 43: This is curious. I would expect some method to get the statistics _out_...

simondsmart: Suggests that the API has different uses than read trivially - and probably needs some documentation commentary at the top.

src/eckit/stats/comparator/ComparatorT.h
```diff
@@ -0,0 +1,64 @@
...
+class Field;
+}
+
+
+namespace eckit::stats::comparator {
+
+
+/// Generic comparison on two Fields
+template <typename STATS>
+class ComparatorT : public Comparator, detail::CounterBinary, STATS {
```
simondsmart, line 34: Because the implementation (and all concrete instantiations) are in the .cc file, I don't think this header file can be successfully included anywhere (well, certainly not if the contents are used).

Perhaps this template class definition would be better at the top of the .cc file...

src/eckit/stats/Comparator.h
```diff
@@ -0,0 +1,97 @@
...
+
+    virtual ~Comparator();
+
+    // -- Operators
+
+    void operator=(const Comparator&) = delete;
+
+    // -- Methods
+
+    virtual std::string execute(const Field&, const Field&) = 0;
```
simondsmart, line 47: std::string is a very non-obvious return type for a Comparator. Unless it is doing something weird.

src/eckit/stats/detail/Counter.cc
```diff
@@ -0,0 +1,166 @@
...
+    countAboveUpperLimit_(0),
+    minIndex_(0),
+    maxIndex_(0),
+    missingValue_(missingValue),
+    lowerLimit_(lowerLimit),
+    upperLimit_(upperLimit),
+    min_(std::numeric_limits<double>::quiet_NaN()),
+    max_(std::numeric_limits<double>::quiet_NaN()),
+    hasMissing_(hasMissing),
+    hasLowerLimit_(lowerLimit_ == lowerLimit_),
```
simondsmart, line 46: This looks odd. I assume it is essentially testing if things are NaN, or similar. But I worry that the equality comparison could be optimised away to true.

src/eckit/stats/method/MethodT.cc
```diff
@@ -0,0 +1,122 @@
...
+        auto stat = s.standardDeviation();
+        *v        = std::isnan(stat) == 0 ? stat : missingValue;
+    }
+
+    field.update(statistics, 0, true);
+}
+
+
+static const MethodBuilder<MethodT<detail::AngleT<double, detail::AngleScale::DEGREE, detail::AngleSpace::ASYMMETRIC>>>
+    __stats1("angle.degree.asymmetric");
```
simondsmart, line 111: Not sure I like the variable names starting with __. Those are supposed to be reserved symbols.

If you really want to ensure non-clashing, you can put them in an anonymous namespace?

src/eckit/stats/util/Mutex.h
```diff
@@ -0,0 +1,58 @@
...
+#include "eckit/thread/Mutex.h"
+#else
+#include <mutex>
+#endif
+
+
+namespace eckit::stats::util {
+
+
+#if defined(ECKIT_GEO_ECKIT_THREADS)
```
simondsmart, line 28: This is a very odd, and somewhat worrying block of code. Why do we need to do special stuff. Why us std::mutex insufficient. And what is the logic here.

If it really is necessary, this needs to be heavily documented as to why (in comments).

tests/stats/stats.cc
```diff
@@ -0,0 +1,176 @@
...
+#include "eckit/config/YAMLConfiguration.h"
+#include "eckit/log/Log.h"
+#include "eckit/stats/FieldStatistics.h"
+#include "eckit/testing/Test.h"
+#include "eckit/types/FloatCompare.h"
+
+
+constexpr double EPS = 1e-6;
+
+#define EXPECTV(a)                                      \
```
simondsmart, line 26: Consider adding these to Test.h

simondsmart (commented):
A few questions in here. Looks good overall.

src/eckit/stats/Comparator.cc
```diff
@@ -0,0 +1,92 @@
...
+
+#include "eckit/exception/Exceptions.h"
+#include "eckit/log/Log.h"
+#include "eckit/stats/util/Mutex.h"
+
+
+namespace eckit::stats {
+
+
+static util::recursive_mutex* local_mutex           = nullptr;
```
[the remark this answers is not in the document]
pmaciel: This is a wrapper around the eckit types that have been superceded by stl; There's a define above that controls this, and I've put it in one place only so it's easy to change.

I called it recursive mutex because that's what the stl calls it, and what the eckit::Mutex really is. But this can of course change, at the moment this is only for controlling access to the factories. On that note, the factories could move to a eckit-based implementation -- which I recently changed to behave correctly for concurrent access, but the change is non-trivial. I think this is what should likely happen first? In any case, happy to update

src/eckit/stats/Comparator.h
```diff
@@ -0,0 +1,97 @@
...
+
+protected:
+    // -- Members
+
+    const Parametrisation& parametrisation1_;
```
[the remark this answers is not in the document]
pmaciel: You're right, I'm not happy about this. Currently the Comparator is a class supporting comparing two statistics (say mean of two fields), but its use is quite specific to pgen-compare fields1 fields2 (which is in fact mir-compare). I probably should take the Comparator hierarchy out? I don't see that MultIO would use this.

src/eckit/stats/Distribution.cc
```diff
@@ -0,0 +1,97 @@
...
+#include "eckit/config/YAMLConfiguration.h"
+#include "eckit/exception/Exceptions.h"
+#include "eckit/log/Log.h"
+#include "eckit/stats/util/Mutex.h"
+
+
+namespace eckit::stats {
+
+
+static util::once_flag once;
```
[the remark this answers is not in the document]
pmaciel: Indeed -- maybe factories refactoring come first though.

src/eckit/stats/Distribution.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace eckit {
+class Parametrisation;
+}
+
+
+namespace eckit::stats {
+
+
+class Distribution {
```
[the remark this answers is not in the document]
pmaciel: You're right -- this is a class to generate noise onto a field according to specific distributions. It might not have a direct use here, but it is important functionality to support statistics testing (incl. downstream packages -- I use this in MIR to generate training data, although this isn't the greatest use). I used this to test the statistics when developing, and I haven't ported those tests as they should be done anew (they aren't useful as is)...

src/eckit/stats/Field.h
```diff
@@ -0,0 +1,38 @@
...
+#include <vector>
+
+
+namespace eckit::stats {
+
+
+using MIRValuesVector = std::vector<double>;
+
+
+class Field {
```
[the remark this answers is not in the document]
pmaciel: Yes you caught on quickly -- this was meant for Mirco to have a look for MultIO first and see what the interface should look like. I've just pushed an update supporting owned and referenced fields, should be much more adequate for MultIO, and fully implemented.

src/eckit/stats/Comparator.h
pmaciel: Indeed this sholuld change -- it is a list of reasons why a comparison failed, and if empty is it a successful comparison. I should bring in more (a lot more) semantics here. Good catch!

src/eckit/stats/detail/Counter.cc
pmaciel: At this level, to support the statistics, they operate on real values and NaN if not assigned/not-definable (there's no "finite missing value"). The assumption here is the NaN support -- and you're right, this can be compiled out. I know that MIR and Atlas will fail in specific places to to similar optimisations. Maybe I should protect this with a static_assert? But the problem is widespread.

src/eckit/stats/util/Mutex.h
pmaciel: This is due to the current threading model supprted in eckit (and the rest of our stacks), to be it's own implementation based on eckit::Mutex. This provides a #define that can change the threading model to stl, but for the stack we should do this in one go -- to me this is a separate discussion. It doesn't work, mixing the models. I've raised this issue over the last few years, several times, but gave up...

src/eckit/stats/method/MethodT.cc
pmaciel: Pushed

src/eckit/stats/Field.h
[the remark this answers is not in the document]
pmaciel: This class is to serve as an argument to calculate field statistics in (for a definition of Field.) You're just too early on the PR as I intended to discuss it beforehand -- but now I've done a proper implementation that can be used in both MIR and MultIO

PR #157, opened 2025-01-14
Constness TCPSocket::read/write methods

mcakircali:


src/eckit/net/TCPSocket.h
```diff
@@ -121,9 +121,9 @@ class TCPSocket {
     int sendBufferSize_    = 0;
 
     // Debug
-    bool debug_;
-    bool newline_;
-    char mode_;
+    mutable bool debug_;
+    mutable bool newline_;
+    mutable char mode_;
```
simondsmart, line 126: This doesn't look right.

mode_ (IIRC) should only be being set in the open() methods (which absolutely should not be const).

mcakircali: it's an unfortunate result of the following (in read/write methods);
```cpp
if (debug_) {
    ...
    if (mode_ != 'r') {
        newline_ = true;
        mode_ = 'r';
    }
    ...
}
```

mcakircali: all those members are used in `if(debug)` blocks, so what's the concern here ?

mcakircali: it should be safer and easier to see after my recent commit

PR #159, opened 2025-01-29
Replaces magic numbers with compile-time variables

mcakircali:
example: seconds in a day would be `eckit::Time::secondsInDay` instead of `86400`

### DIFF
--- a/src/eckit/types/Time.h
+++ b/src/eckit/types/Time.h
@@ -16,6 +16,8 @@
 #ifndef eckit_Time_h
 #define eckit_Time_h
 
+#include <cstdint>
+
 #include "eckit/exception/Exceptions.h"
 #include "eckit/persist/Bless.h"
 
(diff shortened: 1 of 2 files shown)


src/eckit/types/Time.h
```diff
@@ -31,6 +31,12 @@ using Second = double;
 
 class Time {
 
+public: // types
+    static constexpr Second secondsInMinute = 60;
```
simondsmart, line 35: I'm not sure that the type Second is right here. IIRC that is a double - and the count of seconds in * is always integral - and is likely to be used in many integral expressions.

mcakircali: I used `Second` here because `DateTime` and `Time` use `Second (aka double)`

e.g., age calculation cases are `Second`, while STL uses count `time_t (aka long)`

so, `Second` seemed more sensible to me, and no strong opinions.

happy to change if you suggest another type.

danovaro: DateTime and Time use floating points to represent sub-second values.
time_t usually expresses the number of seconds since epoch.
For these constants (positive numbers expressing the relationship between time units, thus a-dimensional), I would prefer an integer value such as an unsigned long

mcakircali: makes sense. there's `useconds_t (aka unsigned int)`, not `unsigned long` ..

any rejection to `std::uint64_t` or to template it `secondsInMinute<double/uint64_t>`?

PR #161, opened 2025-02-05
Fix conditionally declared vars

tmi:
I was having issues compiling eckit because of 
```
/src/eckit/src/eckit/geo/Download.cc: In static member function ‘static eckit::geo::Download::info_type eckit::geo::Download::to_path(const url_type&, const eckit::PathName&, bool)’:
/src/eckit/src/eckit/geo/Download.cc:91:33: error: ‘length’ was not declared in this scope; did you mean ‘Length’?
   91 |     return {static_cast<double>(length), timer.elapsed()};
      |                                 ^~~~~~
      |                                 Length
/src/eckit/src/eckit/geo/Download.cc:91:42: error: ‘timer’ was not declared in this scope; did you mean ‘Timer’?
   91 |     return {static_cast<double>(length), timer.elapsed()};
      |                                          ^~~~~
      |                                          Timer
```

so I'm just moving the variable declaration prior to the `#if` block. Not sure wether this makes sense in case there is no eckit curl, but at least it compiles for me now :)

### DIFF
--- a/cmake/FindCURL.cmake
+++ b/cmake/FindCURL.cmake
@@ -1,0 +1,71 @@
+# (C) Copyright 2025- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+# Try to find CURL
+# See https://curl.se/
+#
+# Defines:
+#  CURL_FOUND - if library was found
+#  CURL_INCLUDE_DIRS - include directories
+#  CURL_LIBRARIES - libraries to link against
+#  CURL_VERSION_STRING - version
+#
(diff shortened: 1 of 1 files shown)


pmaciel:
So @tmi it looks like that curl was found, but the CURL_LIBRARIES variable was set incorreclty (to the directory containing the libs, not the libs themselves.) I didn't find an easy way to ask ecbuild any particular option, nor did I find an existing FindCURL.cmake module -- I really thought one has existed before (?)

The default behaviour of ecbuild is to also package search with modules (which I'm now bundling here) and generally pkgconfig is also present, which this module takes advantage of, and the development version of curl also prepares such a file, which helps.

I'm getting @wdeconinck and @danovaro so we can confirm this is the way forward -- myself I'm against adding more (and more) modules, I didn't see a way out this time. This version works on the hpc, mac + homebrew, docker images (custom dev build) --  I've tried. There are failing tests, but I suspect they've always failed (!) I'm not sure curl (or "EasyCurl") has had working tests, but that's a suspicion only.

pmaciel:
Hi @tmi from our last discussion the code compiles well if CURL found/not found? Or what is the reproducing condition?

PR #163, opened 2025-02-07
Python wheel: add curl and proj dependencies

tmi:
Adding `libcurl` as a dependency for eckit, consists of:
 - yum installing prerequisite libraries (ssl, psl)
 - cloning and building curl (with openssl)
 - adding eckit cmake params
 - including bundled licenses in the wheel
 - calling auditwheel at the end to bundle all libs in (note: we may change this later to manually bundling in, as we had to do with eccodes)

NOTE: the base is not `devel` but `fix/condDeclare` since this one is conditional on it

### DIFF
--- a/python_wrapper/pre-compile.sh
+++ b/python_wrapper/pre-compile.sh
@@ -20,20 +20,27 @@ mkdir -p /tmp/eckit/target/eckit/lib64/
 if [ "$(uname)" != "Darwin" ] ; then
     echo "installing deps for platform $(uname)"
 
-    ## yum-available prereqs
-    yum install -y libpsl-devel openssl-devel 
+    ## curl -- handled by docker container being built with:
+    # dnf instal -y libpsl-devel openssl-devel curl-devel
+    ## this below if the build would need a customization:
+    # git clone https://github.com/curl/curl /src/curl
+    # cd /src/curl
+    # autoreconf -fi
+    # ./configure --prefix /tmp/curl --with-openssl
+    # make -j10 && make install
+    # cd -
 
-    ## curl
-    git clone https://github.com/curl/curl /src/curl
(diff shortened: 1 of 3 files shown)


python_wrapper/pre-compile.sh
```diff
@@ -0,0 +1,54 @@
...
+    # ./configure --prefix /tmp/curl --with-openssl
+    # make -j10 && make install
+    # cd -
+
+    ## proj
+    git clone https://github.com/OSGeo/PROJ /src/proj
+    mkdir /src/proj/build && cd /src/proj/build
+    cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/proj
+    cmake --build . -j8
+    cmake --build . --target install
```
tmi, line 38: @pmaciel I've added the `proj` build here. I've originally started with just `dnf install -y proj-devel`, but that fails to be picked later by ecbuild (and the `PROJ_INCLUDE_DIRS` etc don't seem to be exposed), so I just compile. It allows us to additionally specify whether we want to have eg `libtiff` dependency. And I've replaced the manual curl build with dnf install, that does seem to work ok

tested a linux docker build, seems to work ok. For macos I still need to test a build and possibly add some brew installs, I'll do that once I'll have a "working on a macos laptop" day

pmaciel: I'm quite surprised that curl-devel works, but proj-devel not (proj has better CMake modules support.) So yes this solution works, but we need to revisit this soon to make sure it works as intended. It would be better to fix this now, before the whole workflow is complete (it's slow to improve on working solutions, even if they're costly to maintain.)

There's nothing unusual in the way proj is used in eckit -- if you can reproduce this (the failure of finding proj) let me know and I'll have a look.

PR #166, opened 2025-02-19
Get CUDA backend for eckit_linalg working again

wdeconinck:
- In CMake code, instead of find_package(CUDA), we need to find_package(CUDAToolkit) and link with CUDA::cusparse and CUDA::cublas targets
- Replaced legacy cusparse API with new one.
- The eckit_linalg tests with "cuda" backend pass on our HPC with A100 Nvidia GPUs

Unrelated, a fix for a warning is also in this PR.


pmaciel (approved):

wdeconinck:
Thanks for the rebase!

PR #167, opened 2025-02-19
Add HIP backend for LinearAlgebraSparse to support AMD GPUs

wdeconinck:
This is nearly a copy-paste of the just-merged update of the CUDA backend, with the difference that CUDA symbols are replaced with HIP symbols.

A unit-test with the backend is created and succesfully tested on LUMI with AMD GPUs with cce/17.0.1 and rocm/6.0.3


pmaciel (approved):
Approved just like the CUDA backend :-)

PR #168, opened 2025-02-21
Exception no longer prints to log it set to quiet.

Ozaq:
Exception constructor now respects quiet flag and env variable ECKIT_EXCEPTION_IS_SILENT and no longer logs if any is set.

Fixes: FDB-405

### DIFF
--- a/src/eckit/exception/Exceptions.cc
+++ b/src/eckit/exception/Exceptions.cc
@@ -83,19 +83,7 @@ Exception::Exception(const std::string& w, const CodeLocation& location, bool qu
         Log::error() << "Exception: " << w << " " << location_ << std::endl;
     }
 
-#if 0
-    if (next_) {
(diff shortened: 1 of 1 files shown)


src/eckit/exception/Exceptions.cc
```diff
@@ -83,19 +83,7 @@ Exception::Exception(const std::string& w, const CodeLocation& location, bool qu
...
-    }
-    else
-    {
-        Log::error() << "Exception: stack is empty" << std::endl;
-    }
-#endif
-
     first() = this;
-
-    Log::status() << "** " << w << location_ << std::endl;
```
simondsmart, line 98: This doesn't look to be the right change.

This just removes Log::status() from eckit exception. That is definitely wrong.

Please wrap Log::status() in the quiet check, not remove it.

PR #174, opened 2025-03-12
Add IEC multi-byte unit literals

Ozaq:
Add user defined literals for all 64bit representable power of 2 multi byte units. KiB, MiB, GiB, TiB, PiB and EiB.

Literals are made availabe by including 'utils/Literals.h' and declaring 'using namespace eckit::literals'

Example:

    #include <eckit/utils/Literals.h>
    using namespace eckit::literals;
    const auto bufferSize = 3_KiB;

### DIFF
--- a/src/eckit/utils/Literals.h
+++ b/src/eckit/utils/Literals.h
@@ -1,0 +1,52 @@
+/*
+ * (C) Copyright 2025- ECMWF.
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation nor
+ * does it submit to any jurisdiction.
+ */
+#pragma once
+
+#include <cstdint>
+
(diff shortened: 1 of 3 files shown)


wdeconinck (approved):
Neat!

pmaciel (commented):
@danovaro note that there is a class eckit::Bytes (in eckit/log/Bytes.h) which is now out of sync with this. Two (new) problems:

- change eckit::Bytes unsigned long long to std::uint64_t
- remove eckit::Bytes ZiB/YiB

Since this is merged, I propose @Ozaq to open a new one? It will need to be tested, as the code is much older.

wdeconinck:
> @danovaro note that there is a class eckit::Bytes (in eckit/log/Bytes.h) which is now out of sync with this. Two (new) problems:
> 
> * change eckit::Bytes unsigned long long to std::uint64_t
> * remove eckit::Bytes ZiB/YiB
> 
> Since this is merged, I propose @Ozaq to open a new one? It will need to be tested, as the code is much older.

Is there no automatic conversion from uint64_t to unsigned long long (or size_t)? The eckit::Bytes is just for logging.

Ozaq:
@pmaciel / @wdeconinck I am happy to open a followup PR

PR #175, opened 2025-03-13
Update Bytes.h to use multi-byte unit literals

Ozaq:
This commit also removes Zebi- and Yobibyte static methods from bytes because the result will immediately overflow.

@pmaciel I had a look at checking for wraparound in the added static methods but I decided against it because:
- Unsigned int wraparound is defined behavior in C and C++ (although you can detect it with clangs UB-SAN)
- `Bytes` is used for logging
- Adding a notifying `SafeMultiplication` just for this change seems out of proportion, especially if you want to do this proper you should probably use the available gcc/clang intrinsics `__builtin_uaddll_overflow` et.al. (https://gcc.gnu.org/onlinedocs/gcc-13.3.0/gcc/Integer-Overflow-Builtins.html) BUT ofc the intrinsics are different / not available on INTEL compilers.

Even without explicitly checking the situation is already improved.

Edit: this also now fixes Literals.h not being installed.

### DIFF
--- a/src/eckit/CMakeLists.txt
+++ b/src/eckit/CMakeLists.txt
@@ -679,6 +679,7 @@ list( APPEND eckit_utils_srcs
     utils/Hash.h
     utils/HyperCube.cc
     utils/HyperCube.h
+    utils/Literals.h
     utils/MD5.cc
     utils/MD5.h
     utils/Optional.h
--- a/src/eckit/log/Bytes.h
+++ b/src/eckit/log/Bytes.h
@@ -15,6 +15,8 @@
 #ifndef eckit_log_Bytes_h
 #define eckit_log_Bytes_h
 
+#include "eckit/utils/Literals.h"
+
 #include <iosfwd>
 #include <string>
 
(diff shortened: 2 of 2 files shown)


src/eckit/CMakeLists.txt
```diff
@@ -679,6 +679,7 @@ list( APPEND eckit_utils_srcs
     utils/Hash.h
     utils/HyperCube.cc
     utils/HyperCube.h
+    utils/Literals.h
```
pmaciel, line 682: Oh good!

pmaciel (approved):
I like this PR in that it establishes consistency -- it's not often seen :-) I would have personally implmented these static methods in the .cc rather, but that's just "too much opinion". Great work!

src/eckit/CMakeLists.txt
wdeconinck: Our CI has some gaps 😅

wdeconinck (approved):
❤️

PR #177, opened 2025-03-17
Remove nested SECTION in test_polygon.cc (reported in #71)

wdeconinck:
This PR attempts to address issue #71 .
Nested SECTION scopes in `eckit_test_geometry_polygon` unit test were simply ignored.
By fixing this, it exposes that the unit-test was actually faulty with failed `EXPECT`.

Therefore this PR is currently failing and should be fixed further.
@pmaciel if you don't mind I took the liberty to assign you to see how to fix the failing `EXPECT` as this test is in your expertise and development.

### DIFF
--- a/tests/geometry/test_polygon.cc
+++ b/tests/geometry/test_polygon.cc
@@ -238,7 +238,8 @@ CASE("LonLatPolygon") {
 
         Polygon poly({{lonmin, latmax}, {lonmax, latmax}, {lonmax, latmin}, {lonmin, latmin}, {lonmin, latmax}});
 
-        SECTION("Contains edges") {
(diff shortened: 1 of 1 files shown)


pmaciel:
Indeed, I had forgotten about this and thought it solved (I'm working on the replacement eckit::geo at the moment). There were some issues with the test, mostly because it was a movign target and not because of any mistake. Can you give it another try? I'm happy for you to merge (if all is now good) 👍

pmaciel:
(I just confirmed to have made exactly the same fix in the eckit::geo side!)

pmaciel (approved):

PR #180, opened 2025-03-20
Experiment/libfmt

Ozaq:
Just a draft


tests/exception/test_exceptions.cc
```diff
@@ -61,6 +63,15 @@ CASE("Test ASSERT_MSG does not evaluate message if assertion not triggered") {
     EXPECT_THROWS_AS(ASSERT_MSG(notfail, evaluate_message(notfail)), AssertionFailed);
 }
 
+CASE("Test libfmt") {
+    try {
+        throwException();
```
pmaciel, line 68: Follow this statement with EXPECT(false);

CMakeLists.txt
```diff
@@ -346,6 +346,9 @@ if( NOT ${CMAKE_USE_PTHREADS_INIT} )
     message( FATAL_ERROR "Only pthreads supported - thread library found is [${THREADS_LIBRARIES}]" )
 endif()
 
+### vendored dependencies
+add_subdirectory(third-party)
```
pmaciel, line 350: I've seen these directories typically named "extern" instead (just an observation). In my opinion we should be using git submodules instead, for this purpose, but maybe that's a bridge too far.

Ozaq: I don't disagree :)

PR #182, opened 2025-03-21
Feature/add libfmt

Ozaq:


pmaciel:
(Not a review)

This is an ideal candidate for [git submodules](https://www.atlassian.com/git/tutorials/git-submodule). You can pin to a version, a branch typically main/master. Don't discard the testing....

geier1993:
So the find_dependency stuff is working properly now - also from a bundle.

Now we just need to see why the pyfdb runner with gnu8.5 is using the lib64/ instead of lib/ path

pmaciel:
[In a separate PR](https://github.com/ecmwf/eckit/pull/187) I'm proposing to use git submodules as a way to integrate third-party libraries functionality without copying and with direct access to said libraries versioning. We already have the example of using xxHash, and for libfmt I'm of the opinion of not **copying** code into eckit, especially if that means stripping functionality (documentation, testing).

if you find you have opinions on this please add yourself to review the PR!

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       clang_format: true
       clang_format_ignore: |
         src/eckit/contrib
+        third-party
```
simondsmart, line 34: Why are we creating a second directory here. I don't see libfmt as a dependency being conceptually meaningfully different to xxHash. They are both third party contribs. Please put in the 'original' location.

src/eckit/types/FloatCompare.cc
```diff
@@ -1,35 +1,23 @@
...
-#endif
-
 #include <sys/types.h>
+#include <cmath>
 #include <limits>
 
 #include "eckit/exception/Exceptions.h"
 #include "eckit/types/FloatCompare.h"
 
+using std::fpclassify;
```
simondsmart, line 10: This seems unrelated.

We weren't using the std:: ones as we had issues on some compiler combinations. Not sure I want to couple this change in this context?

simondsmart (commented):
I'm very happy with this, modulo the comment that I don't understand why we are creating essentially a second contrib directory. Please include the code in the same way as other contexts!

src/eckit/types/FloatCompare.cc
geier1993: we included the `eckit/format/Format.h` in `eckit/log/Log.h` - this was then included in `FloatCompare`

Here some macros to math functions were used to avoid having to prefix `std::` ... however as they were declared as macros, they caused a lot of conflicts with the included libtfmt headers.

In that sense using the `using std::function;` approach is much cleaner. I'm not sure if this approach was properly tested before? Moreover we are now using C++17 - I'd be really supprised if there are still compilers complaining about these things. Also the CI passes.

.github/workflows/ci.yml
geier1993: well, it is a third-party library ... same for xxHash. The naming `contribution` is just not appropriate for these because they are no contributions from externals provided just for this purpose.

Also note that in https://github.com/ecmwf/eckit/pull/187  it is to discussed to move them to `third_party` directory as well, and also use git submodules properly...

However, we want to keep the discussion with the git submodules separate. That's why this library is still vendored.

wdeconinck: Semantics... contrib is the directory for external contribution; personally I never heard of the term "vendored" before in this context either.
If instead this is agreed to go to a new directory third-party, then I propose to also move xxHash to this third-party to be consistent at least.

wdeconinck (commented):
I think this is a nice addition, thanks @Ozaq !

I wonder the usefulness of adding these macros currently:
- eckit_format
- eckit_format_cc
- eckit_format_to
- eckit_format_to_cc

Adding macros should be avoided, and I understand these can improve performance from the comments, but I don't think we really need to worry about this. So far we have managed without this. Logging should not be in a critical loop anyways.
If this appears to be a requirement we could add this later?

Then, I don't know why it was chosen to call the new functions `eckit::str_format` (instead of `eckit::format`) and `eckit::str_format_to` (instead of `eckit::format_to`) rather than the similar names in libfmt and C++20 `std::format` and `std::format_to`.
Naming the functions similar will be helpful to port later to C++20.

Is the `ENABLE_FORMAT` procedure similar for libfmt and C++20's format?
Looking at [this blogpost](https://www.cppstories.com/2022/custom-stdformat-cpp20/) it looks a bit different, but I have not investigated further.

geier1993:
> I think this is a nice addition, thanks @Ozaq !
> 
> I wonder the usefulness of adding these macros currently:
> 
> * eckit_format
> * eckit_format_cc
> * eckit_format_to
> * eckit_format_to_cc
> 
> Adding macros should be avoided, and I understand these can improve performance from the comments, but I don't think we really need to worry about this. So far we have managed without this. Logging should not be in a critical loop anyways. If this appears to be a requirement we could add this later?
> 
> Then, I don't know why it was chosen to call the new functions `eckit::str_format` (instead of `eckit::format`) and `eckit::str_format_to` (instead of `eckit::format_to`) rather than the similar names in libfmt and C++20 `std::format` and `std::format_to`. Naming the functions similar will be helpful to port later to C++20.
> 
> Is the `ENABLE_FORMAT` procedure similar for libfmt and C++20's format? Looking at [this blogpost](https://www.cppstories.com/2022/custom-stdformat-cpp20/) it looks a bit different, but I have not investigated further.

The function `eckit::format` already exists to specify manipulators on output streams... It is used at various places... That's why the prefix str_ was added

wdeconinck:
> The function eckit::format already exists to specify manipulators on output streams... It is used at various places... That's why the prefix str_ was added

ugh... bummer.
Could you be more specific which function or places this is used?

Ozaq:
> I think this is a nice addition, thanks @Ozaq !
> 
> I wonder the usefulness of adding these macros currently:
> 
> * eckit_format
> * eckit_format_cc
> * eckit_format_to
> * eckit_format_to_cc
> 
> Adding macros should be avoided, and I understand these can improve performance from the comments, but I don't think we really need to worry about this. So far we have managed without this. Logging should not be in a critical loop anyways. If this appears to be a requirement we could add this later?

About `*_format`:
The primary concern of these macros is the compiletime checking of the format expression, which can only be done on constexpr strings, hence if this was a function the format string argument would no loner be constexpr.

The compiletime checks guard against type mismatches and providing additional arguments to the format call. E.g. format("just a single value: {}", 1 ,2) would not compile, if this is done at runtime an exception would be thrown.

I believe once we have format available this will be used a lot when building messages for exceptions, e.g. "Oh noes x[{}] != y[{}]". Unless all these error code paths are executed they might fail at runtime and instead of throwing the exception we want, we end up with a fmt exception about faulty formatting instead.

I am personally against providing eckit wrappers for fmt but this was what I understood as a requirement. I would consider the effort to manually update from `fmt` to `std` once we are on c++23 to be minor as there are no structural differences and the vast majority of call-sites can be regexreplaced at some point. Further I believe continued use of fmt over std::format beyond c++23 might bring the benefit of continued API evolution over time.

About `*_cc`:

These two macro are primarily concerned with performance and offer a compiletime/codesize vs. runtime allocation behavior tradeoff. 

They are primarily added to highlight that they exist and enable to log from some somewhat performance critical code (apply your own judgment, Its sill not free even if cheaper)

And the PR is still at this point a basis for discussion.

> Then, I don't know why it was chosen to call the new functions `eckit::str_format` (instead of `eckit::format`) and `eckit::str_format_to` (instead of `eckit::format_to`) rather than the similar names in libfmt and C++20 `std::format` and `std::format_to`. Naming the functions similar will be helpful to port later to C++20.

Because it clashes with `Log.h:118 eckit::format(...)` which seems to be used in a lot of our testing.

> Is the `ENABLE_FORMAT` procedure similar for libfmt and C++20's format? Looking at [this blogpost](https://www.cppstories.com/2022/custom-stdformat-cpp20/) it looks a bit different, but I have not investigated further.

What is happening here is that we are using the fmt ostream_formatter (https://fmt.dev/latest/api/#stdostream-support), a formatter that is designed to make types that support the stream oper already printable with format. This fallback does not exist in C++20/23. On migration to std::format (independently of fmt) we would need to add formatters for all types (that we want to print, so a good chance is that this means for all types that already have a ostream oper) as described in the link you added.

tbkr (approved):

tbkr (changes requested):
Check with the other reviewers first.

PR #184, opened 2025-03-26
Add new exception type 'ProtocolError'

Ozaq:
this commit adds a new exception type to signal client server communication erros on the protocol level. I.e. the network is fine but the other side is sending unexpected data or out or order messages.

### DIFF
--- a/src/eckit/exception/Exceptions.h
+++ b/src/eckit/exception/Exceptions.h
@@ -253,6 +253,17 @@ public:
     explicit UnexpectedState(const std::string&, const CodeLocation& = {});
 };
 
+/**
+ * Signals a protocol error in client/server communiation.
+ * Use if the expected communication protocol is broken, i.e. wrong message flow or broken message.
+ * @note This exception is 'silent', i.e. no backtrace is logged.
+ */
+class ProtocolError : public Exception {
+public:
+
+    explicit ProtocolError(const std::string& what, const CodeLocation& = {});
+};
+
 //----------------------------------------------------------------------------------------------------------------------
 
 template <class T>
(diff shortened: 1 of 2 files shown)


src/eckit/exception/Exceptions.h
```diff
@@ -253,6 +253,14 @@ class UnexpectedState : public Exception {
     explicit UnexpectedState(const std::string&, const CodeLocation& = {});
 };
 
+/// Signals a protocol error in client/server communiation.
+/// Use if the expected communication protocol is broken, i.e. wrong message flow or broken message.
+/// @note This exception is 'silent', i.e. no backtrace is logged.
+class ProtocolError : public Exception {
+public:
+    explicit ProtocolError(const std::string& what);
```
simondsmart, line 261: Please accept a location here. All exceptions thrown should (ideally) include the location.

pmaciel: > Please accept a location here. All exceptions thrown should (ideally) include the location.

Hah! that's because I've made that exact change on the previous PR ;-)

pmaciel:
Question, where's the code using this exception? If it isn't in eckit, this exception should be moved there instead (like Atlas and MIR do). Even inside eckit, in (at least the bigger) sub-libraries, the exceptions should be inside those sub-libraries directories (like eckit_geo).

Unless, this is a generic exception that you already have more use cases for (than one)?

Ozaq:
> Question, where's the code using this exception? If it isn't in eckit, this exception should be moved there instead (like Atlas and MIR do). Even inside eckit, in (at least the bigger) sub-libraries, the exceptions should be inside those sub-libraries directories (like eckit_geo).
> 
> Unless, this is a generic exception that you already have more use cases for (than one)?

Hey good question, I intended to use this exception in my work to add Hermes support to the mars client rewrite. So for now there is just one use case, outside of eckit. The protocol will most likely move into metkit. 

I have to admit that I misread the intent of the Exception collection presented in Exception.h. I also would prefer to declare more library specific exception types.

How about ill close this one in favor of a local implementation in either mars-client/metkit and instead add a bit of file level doc to Exception.h explaining essentially what you said?

PR #186, opened 2025-04-01
ECKIT-628 added timeout to RLE encoder

danovaro:
added optional timeout to RLEdecode2 (to be used in mars-server layout compression).
As the timeout expires, we copy the remaining input data and return

### DIFF
--- a/src/eckit/utils/RLE.cc
+++ b/src/eckit/utils/RLE.cc
@@ -126,13 +126,22 @@ stop:
 }
 
 template <class T, class U>
-long long RLEencode2(T first, T last, U output, long long maxLoop, const EncodingClock::duration timelimit) {
-    std::optional<EncodingClock::time_point> deadline{};
-    if (timelimit != EncodingClock::duration::zero()) {
-        deadline = EncodingClock::now() + timelimit;
-    }
-    return RLEencode2timeout(first, last, output, maxLoop, deadline);
+long long RLEencode2(T first, T last, U output, long long maxLoop) {
+    return RLEencode2timeout(first, last, output, maxLoop, std::optional<EncodingClock::time_point>{});
 }
+
+template <class T, class U>
+long long RLEencode2(T first, T last, U output, long long maxLoop, const EncodingClock::duration timelimit) {
+    return RLEencode2timeout(first, last, output, maxLoop,
+                             std::optional<EncodingClock::time_point>{EncodingClock::now() + timelimit});
+}
+
+template <class InputIterator, class OutputIterator>
+long long RLEencode2(InputIterator first, InputIterator last, OutputIterator result, long long maxloop);
+
+template <class InputIterator, class OutputIterator>
+long long RLEencode2(InputIterator first, InputIterator last, OutputIterator result, long long maxloop,
+                     const EncodingClock::duration timelimit);
 
 template <class T, class U>
 void RLEdecode2(T first, T last, U output) {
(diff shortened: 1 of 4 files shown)


src/eckit/utils/RLE.h
```diff
@@ -16,6 +16,7 @@
 
 
 #include <iosfwd>
+#include <chrono>
```
pmaciel, line 20: #include <optional>

src/eckit/utils/RLE.cc
```diff
@@ -11,6 +11,7 @@
 #include <cstddef>
 #include <iterator>
 #include <vector>
+#include <chrono>
```
pmaciel, line 14: Instead of some of these headers, including <chrono>, include RLE.h here

src/eckit/utils/RLE.h
```diff
@@ -26,7 +27,7 @@ namespace eckit {
 class Stream;
 
 template <class InputIterator, class OutputIterator>
```
pmaciel, line 29: I would do something like, before you use steady_clock and in the header:
using EncodingClock = std::chrono::steady_clock;

And use this type instead of explicitly refering to std::chrono::steady_clock. In this overview, I think you can replace every single instance of referring inside std::chrono with this new type.

pmaciel (changes requested):
You're

src/eckit/utils/RLE.h
```diff
@@ -14,19 +14,22 @@
...
+using EncodingClock = std::chrono::steady_clock;
+
 //-----------------------------------------------------------------------------
 
 class Stream;
 
 template <class InputIterator, class OutputIterator>
-long long RLEencode2(InputIterator first, InputIterator last, OutputIterator result, long long maxloop);
+long long RLEencode2(InputIterator first, InputIterator last, OutputIterator result, long long maxloop,
+                     EncodingClock::duration timelimit = EncodingClock::duration::zero());
```
Ozaq, line 32: I like the use of optional in the inner function, I think the interface would be nice if this was also used here, i.e. 
```suggestion
long long RLEencode2(InputIterator first, InputIterator last, OutputIterator result, long long maxloop,
                     const std::optional<EncodingClock::duration>& timelimit = std::nullopt);
```

src/eckit/utils/RLE.cc
```diff
@@ -98,26 +101,36 @@ long long RLEencode2(T first, T last, U output, long long maxLoop) {
...
-        n += RLEencode2(from + k * x, last, output, maxLoop);
+        n += RLEencode2timeout(from, from + x, output, maxLoop, deadline);
+        n += RLEencode2timeout(from + k * x, last, output, maxLoop, deadline);
 
         return n;
     }
 }
 
+template <class T, class U>
+long long RLEencode2(T first, T last, U output, long long maxLoop, EncodingClock::duration timelimit) {
```
Ozaq, line 126: Same as in header
```suggestion
template <class T, class U>
long long RLEencode2(T first, T last, U output, long long maxLoop, const std::optional<EncodingClock::duration>& timelimit) {
```

src/eckit/utils/RLE.h
```diff
@@ -14,19 +14,22 @@
 #ifndef eckit_RLE_h
 #define eckit_RLE_h
 
-
+#include <chrono>
```
Ozaq, line 17: need to add `#include <optional>` here if you apply my suggestion or in cpp if you don't.

src/eckit/utils/RLE.cc
danovaro: we include RLE.cc in RLE.h due to the template, so this change has little impact

PR #188, opened 2025-04-02
Adds safe signed->unsigned, unsigned->signed casts

Ozaq:
Adds new 'into_unsigned()' and 'into_signed' type casts that throw on conversion from a unrepresentable value in the target representation.

### DIFF
--- a/src/eckit/utils/SafeCasts.h
+++ b/src/eckit/utils/SafeCasts.h
@@ -1,0 +1,72 @@
+/*
+ * (C) Copyright 2025- ECMWF.
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation nor
+ * does it submit to any jurisdiction.
+ */
+
+/**
+ * Place functions to allow for save type casting in this file
+ */
+#pragma once
+#include <limits>
+#include <type_traits>
+
+#include "eckit/exception/Exceptions.h"
+
+namespace eckit {
+
+/**
+ *  Casts signed integer into unsigned.
+ *  @param value to cast from.
+ *  @return value cast into, same as before but unsigned type.
+ *  @throws BadCast if used with a negative value.
+ */
+template <typename S, std::enable_if_t<std::is_integral_v<S> && std::is_signed_v<S>, int> = 0>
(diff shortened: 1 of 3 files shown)


simondsmart:
Looks nice. What is the use case you have in mind?

Ozaq:
> Looks nice. What is the use case you have in mind?

I am reading a signed value in Hermes (which should have been a unsigned value), and since I am receiving the value form the network I want to ensure the service is safely converting the value, or raise an error that can be handled gracefully.

wdeconinck (approved):
I think this will be very useful!

wdeconinck:
Note, the unit test fails compilation for rocky-8.6.
https://github.com/ecmwf/eckit/actions/runs/14229901435/job/39878181222?pr=188#step:3:3872

Ozaq:
> i like this PR :-)
> 
> But I would like to raise the discussion, shouldn't this be integrated with eckit::Translator ? It serves a similar purpose, and we'de benefit from integrating this PR's more-natural interface into the more cumbersome Translator operator()(...)

I haven't even looked into translator... (or was particularly aware of its existence) Let me have a look and i'll get back to you.

simondsmart:
Note that we have already added the translate<To>() function which somewhat improves the Translator functionality.

Ozaq:
> i like this PR :-)
> 
> But I would like to raise the discussion, shouldn't this be integrated with eckit::Translator ? It serves a similar purpose, and we'de benefit from integrating this PR's more-natural interface into the more cumbersome Translator operator()(...)

Ok so I had some time to look into the translator. I think this should not become part of the translator. All specializations in Translator.h deal with serialization or deserialization (one of FROM/TO are `std::string`). While it could be integrated into translator just from looking at the interface I don't think it would make for nice semantics.

Ozaq:
Can we move on with this as it is now or do you have specific change requests?

pmaciel:
> Can we move on with this as it is now or do you have specific change requests?

Yes, please merge this with Translator (in the right place) -- otherwise we will start having multiple ways of doing the same thing

wdeconinck:
My guess is miscommunication. I suspect @pmaciel  didn't mean to replace this functionality with eckit::Translator and he doesn't mean that you should be using eckit::Translator. Rather he means to refactor eckit::Translator to make use of this new functionality, if possible.

pmaciel:
Yes indeed - to me the SafeCasts file is fine, it's just  Translator should use this for its existing (and possibly missing) specialisations

Ozaq:
+1 thank ms for the clarification! Will do!

pmaciel (approved):
After a call with @Ozaq we agreed to make a more modern solution, while providing a easy transition.

PR #189, opened 2025-04-22
Python wheel: add testpypi param, push-tag trigger

tmi:
1/ exposes `use_test_pypi` option -- with that set to `true`, the action is safe to run from any branch etc. Later I'll add a trigger to test it automatically on eg a PR to `main`
2/ adds wheel building & uploading invocation to tag push

Note: *needs* https://github.com/ecmwf/reusable-workflows/pull/18 merged


iainrussell (approved):
Looks good to me!

PR #190, opened 2025-05-06
Python wheel: minor fix

tmi:
Minor fix in the bash script for wheel building


wdeconinck (approved):

simondsmart:
This looks like something quite nasty in there is failing in the CI?

PR #195, opened 2025-06-11
Fix formatting

Ozaq:


Ozaq:
> This doesn't look right, that a 'fix formatting' PR is failing the clang-format test...

So we run the CI action on the github event 'pull_request' by default. Additionally we run 'pull_request_target' if there is any label on the PR. Now the problem is that they run in different contexts and also point to different commits. When started with 'pull_request_target' the GITHUB_REF and GITHUB_SHA point to the commit of the target. If this is checked out the code actually tested is the one from develop (with the formatting errors).

wdeconinck (approved):

PR #200, opened 2025-06-20
Feature/eckit codec

pmaciel:
Update to eckit::codec
- record re-compression tool
- support for std::byte scalar and array
- move adapter from namespace std to eckit::codec
- DataType update for uint32_t (saw that change in atlas-io)


pmaciel:
Maybe best to see it as each commit separately

src/eckit/codec/detail/TypeTraits.h
```diff
@@ -14,7 +14,9 @@
 
 #include <type_traits>
 
-#include "eckit/codec/detail/DataType.h"
+#include "eckit/codec/detail/DataType.h"  // enforce visibility
+#include "eckit/codec/types/scalar.h"     // ...
+#include "eckit/codec/types/string.h"     // ...
```
wdeconinck, line 19: I would not add scalar or string here in TypeTraits class.
You could add it just where needed.
Anyway it is already exported in "eckit/codec/codec.h"

PR #207, opened 2025-07-02
fix cmake: bump min ecbuild version

mcakircali:
missing testproperties in ecbuild 3.7

### Contributor Declaration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 cmake_minimum_required( VERSION 3.12 FATAL_ERROR )
 
-find_package( ecbuild 3.7 REQUIRED HINTS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/../ecbuild)
+find_package( ecbuild 3.9.1 REQUIRED HINTS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/../ecbuild)
 
 project( eckit LANGUAGES CXX )
 


mcakircali:
> What has changed that makes 3.9.1 the minimum now?

@simondsmart a test had been added that used TEST_PROPERTIES (see description)

mcakircali:
> > What has changed that makes 3.9.1 the minimum now?
> 
> @simondsmart a test had been added that used TEST_PROPERTIES (see description)

@pmaciel, you added the test, any comments how to progress?

someone needs to merge this or remove the TEST_PROPERTIES as @wdeconinck suggests.

mcakircali:
> @mcakircali, @wdeconinck Yes I agree with all the discussion, I've just pushed a change going closer to CMake syntax. In my opinion there are three problems:
> 
> * TEST_PROPERTIES isn't useful, there's set_tests_properties instead
> * ecbuild's ecbuild_add_test doesn't always add the target, as opposed to CMake's add_test (so any necessary following logic is necessarily different)
> * we always have a mix of old/new deployed versions on the workers but severely limited on moving forward do to significantly older operational platforms, sometimes in spite of modern cmake
> 
> It isn't a good idea to bring more and more features to ecbuild, but rather less. We are really moving away from writing portasble code. In fact if anything, we should be deprecating ecbuild features now that, for the most part, the same functionality or better already exists for CMake versions, including old ones (~2-3 years old). We are almost forcing CMake 2-like syntax because we cannot move, and this is a massive amount of technical debt accumulating.

@pmaciel I understand that you also suggest removing TEST_PROPERTIES, correct me if I'm wrong.

@wdeconinck reverting my commit, can you remove TEST_PROPERTIES and use CMake macros?

I hope we can move this forward.

wdeconinck:
Thanks @mcakircali , in the light of recent commit by @pmaciel  (https://github.com/ecmwf/eckit/commit/5ec356d84d12bdaf6819114f0dc02012815c5696) into develop, we can delay bumping the required version, which in general increases compatibility, bumping just one package to a new version without requiring  to update more packages, to test in isolation.

PR #209, opened 2025-07-02
ECKIT-635 FAM Mock and data structures on FAM via OpenFAM lib

mcakircali:
### Description

Adds a FAM-backed unordered map (plus supporting list/session/URI infrastructure) and an OpenFAM mock implementation to enable development/testing without a real OpenFAM deployment.

Introduces:

- FAM data structures (FamList, FamMap, iterators, entries) and associated naming/URI/session utilities.
- OpenFAM mock (shared-memory based), wired into the cmake configuration.
- suite of OpenFAM/FAM unit tests
- CMake options for OpenFAM/OpenFAM-mock builds

### Contributor Declaration

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-209


simondsmart (changes requested):

src/eckit/io/fam/openfam_mock/FamMockSession.cc
```diff
@@ -0,0 +1,367 @@
...
+    if (shm.creator) {
+        LOG_DEBUG_LIB(LibEcKit) << "Zero-initializing shared memory and mutex.\n";
+        std::memset(mapping_, 0, g_shm_total_size);
+
+        // Initialize process-shared robust mutex.
+        pthread_mutexattr_t attr;
+        ::pthread_mutexattr_init(&attr);
+        ::pthread_mutexattr_setpshared(&attr, PTHREAD_PROCESS_SHARED);
+        ::pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST);
+        const int mrc = ::pthread_mutex_init(&state_->mutex, &attr);
```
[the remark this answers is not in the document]
mcakircali: good catch

src/eckit/io/fam/FamHandle.h
```diff
@@ -0,0 +1,94 @@
...
+
+namespace eckit {
+
+
+//----------------------------------------------------------------------------------------------------------------------
+
+class FamHandle : public DataHandle {
+public:  // methods
+
+    enum class Mode : std::uint8_t {
```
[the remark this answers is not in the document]
mcakircali: no, and not easily movable to DataHandle as it has broader impact.

src/eckit/io/fam/FamList.cc
```diff
@@ -0,0 +1,338 @@
...
+
+auto FamList::descriptor() const -> Descriptor {
+    return {region_.index(), head_.offset(), tail_.offset(), size_.offset(), epoch_.offset()};
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+// iterators
+
+auto FamList::begin() const -> iterator {
+    return region_.proxyObject(FamListNode::getNextOffset(head_));
```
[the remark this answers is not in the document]
mcakircali: creates a FamObject wrapper around an existing object identified by {regionId, offset}
its purpose is to avoid rpc, which happens lookup api

src/eckit/io/fam/FamMapEntry.h
```diff
@@ -0,0 +1,103 @@
...
+
+    using key_type   = FixedString<KeySize>;
+    using value_type = Buffer;
+    using size_type  = std::size_t;
+
+    static constexpr auto key_size = static_cast<size_type>(KeySize);
+
+    /// Encode a key-value pair into a flat buffer for storage in FamList nodes.
+    /// Layout: [key (KeySize bytes)] [value data (length bytes)]
+    static Buffer encode(const key_type& key, const void* data, size_type length) {
```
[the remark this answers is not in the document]
mcakircali: if values are small size, this is more efficient and simpler. 1 memcpy (here) vs +1 fam put (in-situ). otherwise, we need to add a manual fam put path for in-situ.

src/eckit/io/fam/FamPath.cc
```diff
@@ -0,0 +1,117 @@
...
+            return {names[0], ""};
+        case 2:
+            return {names[0], names[1]};
+        default:
+            return {};
+    }
+}
+
+/* ISO Object Identifier Namespace */
+const uuid_t ns_oid = {0x6b, 0xa7, 0xb8, 0x12, 0x9d, 0xad, 0x11, 0xd1, 0x80, 0xb4, 0x00, 0xc0, 0x4f, 0xd4, 0x30, 0xc8};
```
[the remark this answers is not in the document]
mcakircali: ns_oid is the RFC 4122 "OID namespace", which is well-known seed for generating deterministic UUIDs.
we don't "only" use the first element below.
`&ns_oid[0]` is ptr to start, `uuid_generate_sha1` reads all 16 bytes

src/eckit/io/fam/FamSession.cc
```diff
@@ -0,0 +1,420 @@
...
+}  // namespace
+
+//----------------------------------------------------------------------------------------------------------------------
+// SESSION
+
+FamSession::FamSession(std::string name, const net::Endpoint& endpoint) : name_{std::move(name)}, endpoint_{endpoint} {
+    ASSERT(isValidName(name_));
+}
+
+FamSession::~FamSession() {
```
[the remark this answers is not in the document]
mcakircali: this is used in shared ownership model. so, close/finalize is not callable "manually" and also another concern would be any fam invocation after a close (fam_finalize).

tests/io/fam/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+if( eckit_HAVE_OPENFAM )
```
simondsmart: In the main CMakeLists.txt we have:

```
if( eckit_HAVE_OPENFAM_MOCK )
    add_subdirectory( src/eckit/io/fam/openfam_mock )
    set( eckit_HAVE_OPENFAM 1 )
endif()
```

Which suggests that this if/elseif can never reach the second branch.

src/eckit/io/fam/FamSessionManager.h
```diff
@@ -0,0 +1,78 @@
...
+class Endpoint;
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+
+class FamSessionManager {
+public:  // types
+
+    class TestAccessor;
+    using Session = std::shared_ptr<FamSession>;
```
simondsmart, line 41: I don't intuitively understand why this needs to be a shared_ptr.

The purpose of the FamSessionManager is to own/persist the FamSession object for the period of time it is needed for. Not sure what other object needs to maintain shared ownership of it.

I would expect us to have

`FamSession& session(const net::Endpoint& endpoint);`

src/eckit/io/fam/FamSessionManager.h
```diff
@@ -0,0 +1,78 @@
...
+    Session find(const std::string& name);
+
+    // Removes null sessions or older than 30 minutes
+    void cleanup();
+
+private:  // members
+
+    friend class TestAccessor;
+
+    mutable std::recursive_mutex mutex_;
```
simondsmart: Why do we need a recursive_mutex here? That is a touch concerning.

src/eckit/io/fam/FamSessionManager.cc
```diff
@@ -0,0 +1,82 @@
...
+
+
+void FamSessionManager::remove(const std::string& name) {
+    std::lock_guard lock(mutex_);
+    sessions_.remove_if([&name](const auto& session) { return session && session->name() == name; });
+}
+
+void FamSessionManager::cleanup() {
+    // Assumes mutex is held by caller
+    const auto expire_time = std::chrono::system_clock::now() - std::chrono::minutes(30);
```
simondsmart, line 44: Why 30 minutes?

src/eckit/io/fam/FamRegion.h
```diff
@@ -0,0 +1,90 @@
...
+
+    void destroy() const;
+
+    bool exists() const;
+
+    // properties
+
+    fam::index_t index() const;
+
+    fam::size_t size() const;
```
simondsmart, line 47: These are curious. I would expect to be either using std::size_t, or eckit::Length.

src/eckit/io/fam/FamRegion.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace eckit {
+
+//----------------------------------------------------------------------------------------------------------------------
+
+class FamRegion {
+
+public:  // methods
+
+    FamRegion(FamSession& session, FamRegionDescriptor* region);
```
simondsmart, line 37: If we are storing region in a shared pointer, why are we receiving it as a naked pointer? I could understand using shared_from_this if we received region as a reference. But this argument idiom normally implies taking ownership.

What are we going to be _sharing_ the ownership with. That needs to be clearly discussed/commented, as (other than when needed) shared_ptr is a code smell that something is wrong.

src/eckit/io/fam/openfam_mock/FamMockSession.cc
```diff
@@ -0,0 +1,367 @@
...
+struct ShmMapping {
+    int fd{-1};
+    void* mapping{nullptr};
+    bool creator{false};
+};
+
+/// Opens (or creates) the POSIX shm segment, truncates if creator, and maps it.
+ShmMapping openOrCreateShm(const std::string& name) {
+    bool creator = false;
+    auto shmFd   = ::shm_open(name.c_str(), O_CREAT | O_EXCL | O_RDWR, 0666);
```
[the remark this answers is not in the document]
mcakircali: Both the eckit and fdb test helpers append _<PID> to the endpoint before it flows into getShmName(), giving each test process its own POSIX shm segment.

src/eckit/io/fam/FamObject.h
```diff
@@ -0,0 +1,137 @@
...
+
+    void deallocate() const;
+
+    bool exists() const;
+
+    // properties
+
+    fam::index_t regionId() const;
+
+    fam::index_t offset() const;
```
simondsmart, line 71: These fam:: imply that the fam headers need to be public installs, available downstream, rather than this providing a 'proper' API.

src/eckit/io/fam/FamRegionName.cc
```diff
@@ -0,0 +1,74 @@
...
+        return session()->ensureCreateRegion(region_size, region_perm, path().regionName());
+    }
+    return session()->createRegion(region_size, region_perm, path().regionName());
+}
+
+bool FamRegionName::exists() const {
+    try {
+        return lookup().exists();
+    }
+    catch (const NotFound& not_found) {
```
simondsmart, line 58: Normally when we throw (eckit) exceptions stuff gets logged. Does this imply that we will end up with a very polluted output when we do exists() checks, even if debug logging is off?

src/eckit/io/fam/FamRegionName.cc
```diff
@@ -0,0 +1,74 @@
...
+        Log::debug<LibEcKit>() << not_found << '\n';
+    }
+    catch (const PermissionDenied& permission_denied) {
```
simondsmart: Normally PermissionDenied is a rather sharper error than not existing. It doesn't feel like catching this exception and returning not-exists is right.

src/eckit/io/fam/FamSession.cc
```diff
@@ -0,0 +1,428 @@
...
+    invokeFam(&openfam::fam::fam_destroy_region, &region);
+}
+
+void FamSession::destroyRegion(const std::string& region_name) {
+    lookupRegion(region_name).destroy();
+}
+
+FamRegion FamSession::ensureCreateRegion(const fam::size_t region_size, const fam::perm_t region_perm,
+                                         const std::string& region_name) {
+    // Retry loop guards against TOCTOU
```
simondsmart: TOCTOU?

src/eckit/io/fam/FamSession.cc
```diff
@@ -0,0 +1,428 @@
...
+    constexpr int max_retries = 3;
+    for (int attempt = 0; attempt <= max_retries; ++attempt) {
+        try {
+            return createRegion(region_size, region_perm, region_name);
+        }
+        catch (const AlreadyExists&) {
+            try {
+                destroyRegion(region_name);
```
simondsmart, line 206: This feels like a very dangerous function (that creates/destroys regions at will). What is its use?

src/eckit/io/fam/FamTypes.h
```diff
@@ -0,0 +1,62 @@
...
+namespace eckit {
+
+using FamObjectDescriptor = openfam::Fam_Descriptor;
+using FamRegionDescriptor = openfam::Fam_Region_Descriptor;
+
+namespace fam {
+
+constexpr const char* scheme = "fam";
+
+using size_t  = std::uint64_t;
```
simondsmart, line 45: Ah, I may have made a false comment elsewhere. These fam:: types look to be internal to eckit here.

src/eckit/io/fam/FamTypes.h
```diff
@@ -0,0 +1,62 @@
...
+namespace openfam {
+class Fam_Descriptor;
+class Fam_Region_Descriptor;
+}  // namespace openfam
+
+//----------------------------------------------------------------------------------------------------------------------
+
+namespace eckit {
+
+using FamObjectDescriptor = openfam::Fam_Descriptor;
```
simondsmart, line 38: Does this _need_ to leak into the exposed API? Is there anything that the cilent code of these classes needs that needs access to openfam:: internals?

tests/io/fam/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+if( eckit_HAVE_OPENFAM_MOCK )
+    # Mock takes priority: per-process shm isolation via PID-mangled endpoints.
+    list( APPEND _fam_environment
+        "ECKIT_FAM_MOCK_SHM_SIZE=67108864"
+    )
+elseif( eckit_HAVE_OPENFAM )
+    # Real OpenFAM servers running in docker environment.
+    list( APPEND _fam_environment ${_test_environment}
+        "OPENFAM_ROOT=/workspace/cis-rpc_meta-direct_mem-rpc"
+        "ECKIT_FAM_TEST_ENDPOINT=172.26.0.2:8880"
```
simondsmart, line 10: err, what is this ip address. This shouldn't be in publicly released code.

tests/io/fam/test_fam_common.h
```diff
@@ -0,0 +1,137 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+// HELPERS
+
+namespace fam {
+
+// This returns a random number as string, unique per process.
+inline auto random_number() -> std::string {
+    struct timeval tv{};
+    ::gettimeofday(&tv, nullptr);
+    ::srandom(static_cast<unsigned int>(tv.tv_sec + tv.tv_usec + ::getpid()));
```
simondsmart: This reseeds the random number generator each time it is called. This is horrible.

Please use the C++ random number functionality.

tests/io/fam/test_fam_common.h
```diff
@@ -0,0 +1,137 @@
...
+inline const std::string test_endpoint = []() -> std::string {
+    const char* ep = std::getenv("ECKIT_FAM_TEST_ENDPOINT");
+
+    // Real OpenFAM: use the endpoint as-is.
+    if (ep && *ep) {
+        return ep;
+    }
+
+    // Mock OpenFAM: append "_<pid>" to produce a unique POSIX shm path per process.
+    auto base  = std::string("localhost:8880");
```
simondsmart, line 76: Why localhost:8880? Seems odd.
does this work with the mock?

tests/io/fam/test_fam_handle.cc
```diff
@@ -0,0 +1,372 @@
...
+    EXPECT(obj_name.exists());
+
+    // open for write with overwrite — should succeed with smaller length
+    {
+        std::unique_ptr<DataHandle> handle(obj_name.dataHandle(true));
+        handle->openForWrite(32);
+        AutoClose closer(*handle);
+
+        const std::string data = "overwritten";
+        handle->write(data.data(), static_cast<long>(data.size()));
```
simondsmart, line 225: Doesn't test that object has actually been updated

simondsmart (changes requested):

tests/io/fam/CMakeLists.txt
mcakircali: he, why shouldn't some random docker container ip be in code?

tests/io/fam/test_fam_common.h
mcakircali: it just mimics the "real" fam endpoint, no particular reason

tests/io/fam/test_fam_handle.cc
mcakircali: thx, done!

tests/io/fam/test_fam_common.h
mcakircali: good catch, appreciated! though, the choice of word "horrible" is unfortunate.

src/eckit/io/fam/FamTypes.h
mcakircali: well, it's a controlled leak not a full leak.
to be clear, the `type name` is leaked into the API surface, not the headers of openfam.
there are *Descriptor* ctors that are internal but only used as plumbing by FamSession, which is the class with openfam boundary.
we can close it in a follow up PR.

src/eckit/io/fam/FamRegion.h
mcakircali: they're alias to std::uint64_t

src/eckit/io/fam/FamRegion.h
mcakircali: nothing. 
those are the ctors that I mentioned in another comment as internal plumbing used by FamSession. they should have been private. 
raw ptr because we take ownership here, `shared_ptr` because of "copy" semantics, not because shared ownership.
I can enforce move-only and use `unique_ptr` but I don't see what that'd improve in this class.

src/eckit/io/fam/FamRegionName.cc
mcakircali: thx, good catch!

src/eckit/io/fam/FamRegionName.cc
mcakircali: nice "catch", thx

src/eckit/io/fam/FamSessionManager.h
mcakircali: `shared_ptr` mainly because the openfam requires single client object per application per endpoint.
every FamObject/FamRegion object "shares" a single FamSession object (per endpoint).
`shared_ptr` ensures that this single FamSession object is not destroyed before all FamObject/FamRegion objects disappear.
FamSessionManager's main purpose is to manage session objects per endpoint.
thus, we return `shared_ptr` and not ref.

src/eckit/io/fam/FamSessionManager.cc
mcakircali: why not?

src/eckit/io/fam/FamSessionManager.h
mcakircali: must left over during testing/dev.

src/eckit/io/fam/FamSession.cc
mcakircali: overwrite

simondsmart (approved):

PR #210, opened 2025-07-03
Fix compilation of Unzip.cc in legacy conda environment

iainrussell:
### Description
Issue found when building for post-processing conda stack

### Contributor Declaration

### DIFF
--- a/src/eckit/geo/cache/Unzip.cc
+++ b/src/eckit/geo/cache/Unzip.cc
@@ -10,6 +10,9 @@
  */
 
 
+#include <algorithm>
+#include <memory>
+
 #include "eckit/geo/cache/Unzip.h"
 
 #include "eckit/eckit_config.h"


src/eckit/geo/cache/Unzip.cc
```diff
@@ -10,6 +10,9 @@
  */
 
 
+#include <algorithm>
+#include <memory>
```
wdeconinck, line 14: Missing includes, but not related to condo (just coincidence), rather but just generally missing. Perhaps the PR title could be amended.

For consistency, perhaps they should also be guarded by `eckit_HAVE_ZIP`, just like <array> and <fstream> below.

wdeconinck (approved):

pmaciel:
You're right, some are missing, I didn't notice this.

I've put them inside the conditional guards, added one that wasn't spotted so far.

I don't see the use for <algorithm> though?

PR #216, opened 2025-07-22
Support log and syslog over tcp 

mcakircali:
### Description

add support for logging targets over TCP

### Contributor Declaration


src/eckit/log/SysLogTCPTarget.cc
```diff
@@ -0,0 +1,39 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+
+SysLogTCPTarget::SysLogTCPTarget(net::TCPSocket& socket, SysLog log) : TCPTarget(socket), log_(std::move(log)) {}
+
+void SysLogTCPTarget::msgid(int msg_id) {
+    log_.msgid(msg_id);
+}
+
+void SysLogTCPTarget::write(const char* start, const char* end) {
+    log_.message({start, end});
```
simondsmart, line 29: Do we need to consider the behaviour of newlines here, or is that taken care of elsewhere in the logging infrastructure?

src/eckit/log/SysLogTCPTarget.cc
```diff
@@ -0,0 +1,39 @@
...
+#include "eckit/log/SysLog.h"
+#include "eckit/net/TCPSocket.h"
+
+namespace eckit {
+
+//----------------------------------------------------------------------------------------------------------------------
+
+SysLogTCPTarget::SysLogTCPTarget(net::TCPSocket& socket, SysLog log) : TCPTarget(socket), log_(std::move(log)) {}
+
+void SysLogTCPTarget::msgid(int msg_id) {
```
simondsmart, line 24: I'm not sure that we need this magic function.

src/eckit/log/SysLogTCPTarget.h
```diff
@@ -0,0 +1,64 @@
...
+namespace net {
+class TCPSocket;
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+
+/// sends SysLog (RFC 5424 format) logs over TCP
+class SysLogTCPTarget : public TCPTarget {
+
+    /// Example SysLog:
```
simondsmart, line 29: Not sure that this example is correct.

For initialising the Target, the content of "message" is irrelevant.

And for sending the messages, the user no longer needs to do this.

src/eckit/log/SysLog.h
```diff
@@ -70,7 +71,7 @@ class SysLog {
     static char nilvalue() { return '-'; }
 
 
-    SysLog(const std::string& msg, int msgid = 0, Facility f = SysLog::User, Severity s = SysLog::Info);
+    explicit SysLog(std::string msg, int msgid = 0, Facility f = SysLog::User, Severity s = SysLog::Info);
```
simondsmart, line 74: Given that we are going to be using this in a mode where the messages is set incrementally, it may make sense to have a constructor that does not require a message to create it initially.

simondsmart (commented):
Pretty happy with this. Just a couple of implementation questions.

src/eckit/log/SysLog.h
mcakircali: yes, it's annoyed me too. I tend to enlarge the PRs.. this time it's opposite

src/eckit/log/SysLogTCPTarget.cc
mcakircali: it's handled on remote side

src/eckit/log/SysLogTCPTarget.cc
mcakircali: left for request id, which we don't have yet

src/eckit/log/SysLogTCPTarget.h
mcakircali: side affect of different mode that we use this for

PR #218, opened 2025-09-03
c++20 - first draft

danovaro:
### Contributor Declaration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,13 +9,13 @@
 ########################################################################################################################
 
 
-cmake_minimum_required( VERSION 3.12 FATAL_ERROR )
+cmake_minimum_required( VERSION 3.18 FATAL_ERROR )
 
 find_package( ecbuild 3.11 REQUIRED HINTS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/../ecbuild)
 
 project( eckit LANGUAGES CXX )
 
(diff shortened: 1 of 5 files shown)


pgeier:
How likely is it that we move the whole stack to C++20? When it is possible to do for eckit, it must be possible for the whole stack?

wdeconinck:
> How likely is it that we move the whole stack to C++20? When it is possible to do for eckit, it must be possible for the whole stack?

If eckit requires c++20 then naturally everything depending on eckit (so nearly every of our C++ packages) will have c++20 enabled.

pgeier:
I could really make use of `concepts`, `constexpr std::strings`, `designated initializers`, `ranges` & `views` (e.g. replacing the Tokenizer).

wdeconinck:
@tlmquintino , @simondsmart , @Ozaq , relating to the entire software stack (but starting with eckit), do we have a proposed time line, (or could we have one if not), on when we drop support for certain compilers / versions? Could this be made available publicly on a webpage?
Similarly such page should state which compilers/version we currently guarantee due to CI-testing, and then compilers which we aim to support to best efforts but cannot always test, either because we didn't yet do, or because we don't have access (e.g. we should aim to support HPE/Cray, AMD/AOCC, NEC, NAG (Fortran), ...)

We should also involve such timeline for the version of CMake, and standard version of C and C++.

Such timeline should help us to deprecate CI runners and upgrade to new ones. Similar to how Github's own runners get deprecated over time.

PR #219, opened 2025-09-08
Restore compatibility with ecbuild-3.7

wdeconinck:
### Description

In a update for IFS (diab_CY49R1_for_49R2_v9_hackathon_mtg2sc), I believe for ERA6,
a metkit branch "feature/mtg2_merge" is needed/tested. This recently got updated to require eckit 1.31.4,
which in turn requires ecbuild 3.11 as of recent.

With this setup, the particular ifs-bundle does not build, without more changes to other projects.
It seems ecbuild 3.7 compatibility can still be maintained within eckit,
and the particular IFS configuration was able to successfully build with this PR.

FYI, @andrewcbennett @towil1

### Contributor Declaration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 cmake_minimum_required( VERSION 3.12 FATAL_ERROR )
 
-find_package( ecbuild 3.11 REQUIRED HINTS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/../ecbuild)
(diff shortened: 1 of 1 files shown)


pmaciel:
Looking the differences up from 3.7 to 3.11, I can only spot that, in bundles, project-specific options (project_ENABLE_option) are better handled, and this something we explicitly do in the synchronised releases, although I doubt we do this for the production stacks.

The other differences I note are improved handling of GNU-specific cases, a more flexible FFTW packing finding and better testing. I don't think the production stack is benefiting from these changes, but I might tbe wrong here, especially on platforms I don't work with (DestinE).

PR #222, opened 2025-09-18
eckit::geo fix linking of libzip

iainrussell:
### Description
There is a problem in linking the geo lib of eckit when the build environment contains libzip but not shapelib (e.g. conda-forge). This fixes that issue.

### Contributor Declaration

### DIFF
--- a/src/eckit/geo/CMakeLists.txt
+++ b/src/eckit/geo/CMakeLists.txt
@@ -60,8 +60,6 @@ list(APPEND eckit_geo_srcs
     cache/Download.h
     cache/MemoryCache.cc
     cache/MemoryCache.h
-    cache/Unzip.cc
-    cache/Unzip.h
     container/PointsContainer.cc
     container/PointsContainer.h
     figure/Earth.cc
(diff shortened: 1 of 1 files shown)


pmaciel:
Oh that's a surprise -- i thought this was protected by the top-level CMakeLists.txt

ecbuild_add_option( FEATURE GEO_AREA_SHAPEFILE
                    DEFAULT ON
                    CONDITION eckit_HAVE_ECKIT_GEO AND eckit_HAVE_ZIP
                    REQUIRED_PACKAGES shapelib
                    DESCRIPTION "eckit::geo geometry library support for shapefiles" )


where the option shouldn't be enabled unless eckit_HAVE_ECKIT_GEO AND eckit_HAVE_ZIP are on. Maybe this might (also) not be the right way to protect it?

iainrussell:
Ah, that wasn't the problem - the problem is if you have zip but not shape - then the shape feature is disabled, but Unzip.cc is still compiled. But we do not link with libzip unless shape has been enabled, and then we get a linker problem because Unzip.cc needs to link with libzip, but it has not been told to link with it (because libzip is only added to the linker if shape was enabled).

pmaciel:
Can you maybe try with this version?
https://github.com/ecmwf/eckit/tree/fix/geo-ziplib-alternate

This one does not compile cache/Unzip if there's no shapefile functionality either -- in line with the top-level CMakeLists.txt

iainrussell:
Thanks Pedro, on the face of it, that looks better if libzip is only used in conjunction with shapefile support

PR #225, opened 2025-10-02
Type to string conversion utilites

pgeier:
### Description

This PR provides utilities to convert arbitrary types to strings by calling `typeToString<T>()`. 
This utility is useful to create detailed error (and log) messages in templated code or to generate 
a set of specialized error messages consistently.

A typical usecase is type dispatching with `std::variant`:

```
using Value = std::variant<bool, long, double, std::string>;

template<typename T>
T getValue(const Value& var) {
   if (std::holds_alternative<T>(var)) {
      return std::get<T>(var);   
   }
   std::ostringstream oss;
   oss << "Expected type " << typeToString<T>() << " but contained ";
   oss << std::visit([](const auto& v){ return typeToString<std::decay_t<decltype(v)>>(); }, var);
   throw std::runtime_error(oss.str());
};

```

### Contributor Declaration

### DIFF
--- a/src/eckit/utils/TypeToString.h
+++ b/src/eckit/utils/TypeToString.h
@@ -1,0 +1,142 @@
+/*
+ * (C) Copyright 2025- ECMWF.
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation nor
+ * does it submit to any jurisdiction.
+ */
+
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <optional>
+#include <string>
+#include <tuple>
+#include <type_traits>
+#include <variant>
+#include <vector>
+
+#include <string_view>
+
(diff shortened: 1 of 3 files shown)


tbkr:
This looks good to me, as soon as the clang format is done.

pgeier:
There is alread a `demangle` approach used: https://github.com/ecmwf/eckit/blob/develop/src/eckit/codec/detail/demangle.cc

If that approach is accepted to be safer, I'm fine with that. It just returns values like "i" for an "int" on some compilers.

wdeconinck:
You are right, the returned string from `typeid(x).name()` is cxx-mangled, I had forgotten this.
Either way, using these _PRETTY_FUNCTION_ or the `detail::demangle` function is fine by me.

This is a nice feature to have btw. 👌

pgeier:
> You are right, the returned string from `typeid(x).name()` is cxx-mangled, I had forgotten this. Either way, using these _PRETTY_FUNCTION_ or the `detail::demangle` function is fine by me.
> 
> This is a nice feature to have btw. 👌

Uups, we commented coincidentally. Do you feel there is a "strong" need for it? or maybe just a more simplified version without an additional wrapper?

wdeconinck:
There's no strong need for it. Indeed "nice-to-have" but custom use case in multio is probably good idea for now.

PR #229, opened 2025-10-16
Python: enable proj in wheels

tmi:
Enables proc for the python wheel -- relying on a brew install on mac, building custom on linux


Ozaq (approved):

Ozaq (changes requested):
I think this is better suited to be included in cxx-dependencies

PR #279, opened 2026-02-26
Cleanup

Ozaq:
### Description

commit b72c5fc13a2d3b5aa0761e073db2f9f902dad255
Author: Kai Kratz <Kai.Kratz@ecmwf.int>
Date:   Thu Feb 26 20:20:43 2026 +0000

    Add compile-time test to prevent header regressions

    The IWYU cleanup made all public headers self-contained, but nothing
    prevents regressions. Add a CMake test that compiles each header in
    isolation — if a future change breaks self-containedness, the build
    fails immediately pointing to the offending header.

    Gated behind -DENABLE_TEST_SELF_CONTAINED_HEADERS=ON to avoid the
    extra compile cost on regular builds.

    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

commit 94057168c1878df2a16330309d9df44cc4a905b5
Author: Kai Kratz <Kai.Kratz@ecmwf.int>
Date:   Thu Feb 26 20:20:42 2026 +0000

    Ensure headers are self-contained after IWYU cleanup

    Downstream projects include eckit headers in arbitrary combinations
    and orders. A header that silently depends on a transitive include
    will break when unrelated changes elsewhere shuffle include graphs.

    Run IWYU 0.25 across the codebase to add missing direct includes
    and remove unnecessary ones, so every header compiles on its own.

    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

commit e82c6c4bd1932b596d30c77067695bd1d6b33c22
Author: Kai Kratz <Kai.Kratz@ecmwf.int>
Date:   Thu Feb 26 20:20:42 2026 +0000

    Unify include guards to #pragma once

    Traditional #ifndef guards are verbose, error-prone (copy-paste of
    wrong macro names), and add noise to diffs. #pragma once is supported
    by all compilers we target and eliminates these issues.

    - Convert 489 headers from #ifndef guards to #pragma once
    - Add #pragma once to 5 headers that had no guard at all
    - Skip src/eckit/contrib/ (third-party code)

    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### Contributor Declaration

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-279


src/eckit/cmd/UserInput.cc
```diff
@@ -1,12 +1,13 @@
-#include <termios.h>
-#include <unistd.h>
+#include "eckit/cmd/UserInput.h"
```
marcosbento: Please add the typical LICENSE header to this file. Also, check if a similar header is missing on any other file.

marcosbento (changes requested):

Ozaq:
> In my experience, IWYU is good but does generally require manual intervention -- so it is only an initial (good) guess.
> 
> Related, can you also replace all the header include guards #ifndef ... to #pragma once?

IWYU is only used to update the files, no automation, the self containment tests uses a simple compilation per header (thats why its an extra feature as it doubles the compile time)

`#pragma once` is so fixed by this PR, if you still see old include guards here its a bug :)

src/eckit/cmd/UserInput.cc
Ozaq: *Sigh* something I did not check for, will do!

src/eckit/bases/Loader.h
```diff
@@ -11,11 +11,7 @@
 // File Loader.h
 // Baudouin Raoult - ECMWF Jul 96
 
-#ifndef eckit_Loader_h
-#define eckit_Loader_h
-
-#include "eckit/eckit.h"
-
+#pragma once
```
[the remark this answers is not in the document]
Ozaq: hmm don't want to bother with it because I would need to code a manual check for it. clang-format does not support it and without a check it will deteriorate quickly again. 

Although I agree it would look nicer.

PR #287, opened 2026-04-08
Hotfix/2.0.8

pmaciel:
Project packaging sometimes uses distributions to configure & build that have different default path locations to the target platforms. Specifically the (eckit dependency) CURL default CA bundle path does not exist on the distributed Python packages supporting Earthkit, which is intended for a very diverse audience. On failing to find the, CURL cannot verify the host/target connection to do SSL-based transfers.

Here proposed, if there are no (related) environment-controlling variables nor the internal default leads to an existing path, a series of path searches are attempted based on common, expected dirname and basename combinations, local path (".") is avoided for security reasons.

Unfortunately, no test is added because handling these paths are commonly permission-limited.


pmaciel:
In researching the MeteoSwiss failure a bit further and looking into CURL, the failure they see can be

1. TLS handshake fails (that could happen from failing to find a CA bundle, which this PR addresses)
2. DNS resolution failure
3. connection refused or timeout

So while this PR focuses on one of the two failures, it could solve both -- @iainrussell maybe it would be good to tell them directly to test this branch (somehow!), we don't have a testing environment

Ozaq:
> In researching the MeteoSwiss failure a bit further and looking into CURL, the failure they see can be
> 
> 1. TLS handshake fails (that could happen from failing to find a CA bundle, which this PR addresses)
> 2. DNS resolution failure
> 3. connection refused or timeout
> 
> So while this PR focuses on one of the two failures, it could solve both -- @iainrussell maybe it would be good to tell them directly to test this branch (somehow!), we don't have a testing environment

This indicates that we need to improve the error message and communicate the underlying error.

src/eckit/io/EasyCURL.cc
```diff
@@ -449,6 +450,62 @@ class CURLHandle : public eckit::Counted {
...
+            };
+
+            const char* dirs[]{
+                "/etc/pki/tls/certs",
+                "/etc/ssl",
+                "/etc/ssl/certs",
+                "/etc/pki/ca-trust/extracted/pem",
+                "/usr/local/share/certs",
+                "/usr/local/etc/openssl@3",
+                "/opt/homebrew/etc/openssl@3",
```
Ozaq, line 477: ca-certs form the base of trust for TLS secured communication and thus the paths are set by the distribution. By picking up certs from additional locations we increase the possible attack surface. 

Additionally this is a problem with a "well known" solution, i.e. setting the above mentioned env variables.

Ozaq (changes requested):
I think a much better approach is to follow what netcdf and Fiona are doing:

Both python projects solve this issue by using the 'certifi' package, essentially a ca store installed through pip.

They then go on and check if SSL_CERT_FILE or CURL_CA_BUNDLE is set (in earthkit) and configure eckit with the location (this code needs to be added.

This will change the behavior to: always use certs from certifi unless user overrides with env variables.

Pulling up the selection logic into earth kit is imo the right choice because this is the place where we are guratueed to know the location of a valid store (certifi)

PR #294, opened 2026-06-02
Feature/adr004

pmaciel:
### Description

This PR fixes the behaviour of the EasyCURL interfacing to CURL under unusual conditions: presence of firewalls, url redirections, via python bindings and integration with python-based SSL configuration-specific modules (certifi), and similar.

This PR:
- fixes handling of redirects under proxy/tunnel flows
- fixes parsing of GET response with colons (which is common)
- eckit python bindings report version (eg. to check under testing)
- eckit python bindings use certifi to point to a valid CA bundle certificate for SSL authentication
- adds multi-threaded safety to (some) EasyCURL functionality
- maps options by name to internal CURL options, adding a global "setopt" method to the behaviour of future EasyCURL handles

The options mapping might be unnecessary (shall I remove them?) because the fixes above have been proven sufficient to use within the CSCS MeteoSwiss environment which prompted these fixes -- that is, they don't need to use the extra options this PR allows; The use of CURLOPT_VERBOSE is really useful and what allowed a quick diagnostic, so I would like to keep (some of) these options in, but mayeb we could remove most of the others -- don't know.

### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md).
 

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-294


python/eckit/src/eckit/geo/__init__.py
```diff
@@ -11,5 +11,10 @@
...
 findlibs.load("eckit_geo", "eckitlib")
 
-
+from eckit._certifi import configure_ca_bundle_from_certifi
 from eckit.geo._eckit_geo import *
+
+configure_ca_bundle_from_certifi()
+
+__version__ = version()
+__git_sha1__ = git_sha1()
```
tmi, line 20: where do the version/git sha come from? For the eckit, I see they are declared in the pyx, but for the eckit/geo, they don't seem to be? Or do I just misunderstand it?

python/eckit/src/eckit/__init__.py
```diff
@@ -12,4 +12,10 @@
 findlibs.load("eckit")
 
 
+from eckit._certifi import configure_ca_bundle_from_certifi
 from eckit._eckit import *
+
+configure_ca_bundle_from_certifi()
+
+__version__ = version()
```
tmi: I wonder however whether this is right -- the `__version__` is a standard python attr which should be pretty close to `importlib.metadata.version("eckit")`, ie, the version of the wheel that provided this library. Not all python libraries have an exact match (for example, for `torch`, you get the cuda build tag in one but not in other), but an actual mismatch can cause havoc -- eg in forecastbox, we install packages at runtime and then query the `__version__` to see whether the install has gone right, and the package declaring itself incorrectly will always crash down the system

wouldn't it be better to use a _different_ field for this, like `__lib_version__`, if its just for our purposes? And keep `__version__` corresponding to the wheel's version?

ultimately, it should always be true that
```
eckit.wheel.version = eckit.__version__ = eckitlib.wheel.version = eckitlib.__version__ = libeckit.so version
```
but ideally we dont mask one with the other, ie, always provide all of them for a comparison/troubleshooting. And there are cases where it would not necessarily match, like editable builds, findlibs customizations, etc

python/eckit/src/eckit/geo/__init__.py
pmaciel: Sorry this was missed from the original push (because it's made from cherry picked commits from another branch), I've made this right now. Indeed, I'm treating version numbers separately, that eckit can have a version and eckit::geo another (internally, it get's the same version from cmake, but the interface is right.) That is to allow an easy migration when components can be kept separate from the main eckit library, as it should (eventually).

python/eckit/src/eckit/__init__.py
pmaciel: Here I'm in the dark, I'm only trying to mimick the behaviour I see (but I don't know any better). I just wanted to confirm which version was being used in separate venv's I use to test. 

I have no issue in setting __lib_version__ instead (and sha1 I guess), please advise -- I think this is important to get right and I'm not really qualified to know the best practice :-)

tmi: yeah, lets use the `__lib_version__` or smth like that please, to not collide with the expectations; the `__git_sha1__` can stay as is

tmi: or you could set `__eckitlib_version__` and `__eckitgeolib_version__` both on the top level init -- I think its a bit unusual to have `__version__`-like attributes in submodules, because they are package-level attributes. But I may be wrong with this. Whatever you find more convenient, this is just a suggestion

pmaciel: I'll do the __lib_version__ one as I would really like to see eckit being split into separate modules (meaning, separate repositories and versioning for the modules); I see what you mean and I don't mean to disagree (I don't!) But what we're building in eckit is the supporting stack for all the centre's services in a modular way (or, should be). The discussion hasn't happened yet in this regard and that's to our detriment... but this is one of the things that need to happen, module isolation and versioning.

tmi: right. Thinking of it, there are actually very many cases where even `__version__` is declared in a nested module -- the namespace modules, like `earthkit`, where `earthkit.data` or `earthkit.plots` each comes from a different repo, so `earthkit.__version__` itself makes no sense. And I guess `eckit` could face a similar fate -- in which case the nested `__lib_version__` does make sense

I myself don't like these namespace modules, because they dont play along that nicely with some python tools, but I can understand the reasoning for picking it

tmi (approved):

pmaciel:
I've just made a change to the branch with cleaner commits, added the missing version()/gitsha1() methods and removed unusual CURL configuration options to keep the code and logic simple. It was difficult to extract focused changes from a much broader set of changes. Have a look...

PR #297, opened 2026-06-09
[python] update eckit building

tmi:
Whats happening here:
1. fixes the symlink to be non-recursive and point out to the correct files
2. deletes the obsolete build_chain
3. introduces a simplified setup.py, so that
  a) the right version of the wheel is determined
  b) the right eckitlib dependency is propagated
  c) the .so files are included
  d) the wheel is marked as binary despite not being compiled from python's pov
4. updates the buildconfig to enable python and set correct install dirs

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-297


tmi:
wheels tested with https://github.com/ecmwf/python-develop-bundle/actions/runs/27208066364

seems to work:
```
$ uv pip install --prerelease=allow -i https://test.pypi.org/simple/ eckit==2.0.8.dev91
Using Python 3.14.3 environment at: venv
Resolved 9 packages in 167ms
Prepared 1 package in 362ms
Installed 2 packages in 13ms
 + eckit==2.0.8.dev91
 + eckitlib==2.0.8.dev91

$ uv run python
Python 3.14.3 (main, Feb  3 2026, 22:52:18) [Clang 21.1.4 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import eckit
>>>
```

PR #300, opened 2026-06-18
ECKIT-682: Use ~/run if available for mutable state

simondsmart:
### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md).
 

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-300

### DIFF
--- a/src/eckit/io/cluster/ClusterDisks.cc
+++ b/src/eckit/io/cluster/ClusterDisks.cc
@@ -216,7 +216,7 @@ static DiskArray* clusterDisks = nullptr;
 static pthread_once_t once     = PTHREAD_ONCE_INIT;
 
 static void diskarray_init() {
-    LocalPathName path("~/etc/cluster/disks");  // avoid recursion...
+    LocalPathName path(LocalPathName("~/run").exists() ? "~/run/cluster/disks" : "~/etc/cluster/disks");  // avoid recursion...
 
     size_t disksArraySize = Resource<size_t>("disksArraySize", 10240);
 
--- a/src/eckit/io/cluster/ClusterDisks.cc
+++ b/src/eckit/io/cluster/ClusterDisks.cc
@@ -25,6 +25,7 @@
 #include "eckit/io/cluster/NodeInfo.h"
 #include "eckit/log/JSON.h"
 #include "eckit/memory/Zero.h"
+#include "eckit/runtime/Main.h"
 #include "eckit/system/SystemInfo.h"
 #include "eckit/thread/AutoLock.h"
 #include "eckit/utils/Tokenizer.h"
@@ -216,7 +217,7 @@ static DiskArray* clusterDisks = nullptr;
 static pthread_once_t once     = PTHREAD_ONCE_INIT;
 
 static void diskarray_init() {
-    LocalPathName path(LocalPathName("~/run").exists() ? "~/run/cluster/disks" : "~/etc/cluster/disks");  // avoid recursion...
+    LocalPathName path(Main::instance().rundir() / "cluster/disks");
 
     size_t disksArraySize = Resource<size_t>("disksArraySize", 10240);
 
--- a/src/eckit/io/cluster/ClusterNodes.cc
+++ b/src/eckit/io/cluster/ClusterNodes.cc
@@ -16,6 +16,7 @@
 #include "eckit/config/EtcTable.h"
 #include "eckit/config/Resource.h"
 #include "eckit/container/MappedArray.h"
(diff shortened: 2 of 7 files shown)


src/eckit/io/cluster/ClusterDisks.cc
```diff
@@ -216,7 +216,7 @@ static DiskArray* clusterDisks = nullptr;
 static pthread_once_t once     = PTHREAD_ONCE_INIT;
 
 static void diskarray_init() {
-    LocalPathName path("~/etc/cluster/disks");  // avoid recursion...
+    LocalPathName path(LocalPathName("~/run").exists() ? "~/run/cluster/disks" : "~/etc/cluster/disks");  // avoid recursion...
```
marcosbento: This code makes lot of assumptions about what `~/run` is and contains.
Of course, the issue already existed in the previous version, but it gets slightly more convoluted by the new conditional.

To make it slightly more readable, the code for selecting the prefix and the path itself could be extracted/commented to explain the reasoning:

```
static const std::string& generic_select_prefix()
{
    // Prefer `~/run` over `~/etc`, so that `~/etc` can be mounted read-only in a container.
    static const char* default_prefix = "~/etc";
    static const char* preferred_prefix = "~/run";

    if (LocalPathName run(preferred_prefix); run.exists() && run.isDir())
    {
        return preferred_prefix;
    } else
    {
        return default_prefix;
    }
}

static auto generic_select_path(const std::string& suffix)
{
    return generic_select_prefix() + suffix;

    // n.b. We could also perform some validation/setup...
    // For example, it could `mkdir -p ...`
}

static void diskarray_init() {
    auto path = generic_select_path("/cluster/disks");  // avoid recursion...

...
```

src/eckit/io/cluster/ClusterNodes.cc
```diff
@@ -242,7 +243,7 @@ static pthread_once_t once = PTHREAD_ONCE_INIT;
 static std::set<std::string> offsiteNodes_;
 
 static void init() {
-    nodeArray = new NodeArray("~/etc/cluster/nodes", 1024);
+    nodeArray = new NodeArray(LocalPathName("~/run").exists() ? "~/run/cluster/nodes" : "~/etc/cluster/nodes", 1024);
```
marcosbento: And here, the code could reuse the prefix selection logic...

```
nodeArray = new NodeArray(generic_select_path("/cluster/nodes"), 1024);
```

src/eckit/runtime/Monitor.cc
```diff
@@ -83,7 +83,10 @@ static pthread_once_t once          = PTHREAD_ONCE_INIT;
 
 static void taskarray_init(void) {
 
-    LocalPathName monitorPath(Resource<std::string>("monitorPath", "~/etc/monitor"));
+    LocalPathName monitorPath(Resource<std::string>("monitorPath", ""));
+    if (monitorPath == "") {
+        monitorPath = LocalPathName("~/run").exists() ? "~/run/monitor" : "~/etc/monitor";
```
marcosbento: And here, again...

```
 monitorPath = generic_select_path("/monitor");
```

marcosbento (approved):
The changes won't break anything (that wasn't already broken before)!
But this could be a missed improvement opportunity...

marcosbento:
A quick grep shows that there are a few other `~/etc` being used, which don't get the same treatment. For example, the ResourceMgr seems to be loading configs from `~/etc/config` and FileSpace uses `~/etc/disks`.

Should the proposed changes (i.e. the fact that now we are using `~/run` whenever it exists) consider how this affects the other uses of `~/etc`?

simondsmart:
> A quick grep shows that there are a few other `~/etc` being used, which don't get the same treatment. For example, the ResourceMgr seems to be loading configs from `~/etc/config` and FileSpace uses `~/etc/disks`.
> 
> Should the proposed changes (i.e. the fact that now we are using `~/run` whenever it exists) consider how this affects the other uses of `~/etc`?

That was obviously insufficiently clear. etc is generally the right path.

The issue is that etc is for config - and some files (monitor, state files for cluster) are runtime mutable and maintain the running state of the service. And, if something goes wrong, can be nuked and regenerated on a fresh run. Which is completely different behaviour than for config.

I want to be able to mount config read-only in a container. So this change is to pull non-config out of etc.

I take your point about generalisation. I've taken a slightly different approach, but have updated the PR.

PR #335, opened 2026-09-07
[python] bugfix wrong proj path

tmi:
Hotfixing since the recent switch of cxx deps location on macos runners broke this

🌦️ >> Documentation << 🌦️
https://sites.ecmwf.int/docs/dev-section/eckit/pull-requests/PR-335


tmi:
https://github.com/ecmwf/python-develop-bundle/actions/runs/34096826357/job/101662295459 a successful run (at least on the arm64)

